PR #26, opened 2020-04-15
Use Unauthorized instead of AccessForbidden for 401 response

XaosSintez:


ashkarpetin (approved):

slavcodev (approved):
LGTM. I wonder is `AccessForbidden` still used? There are two similar definitions `AccessForbidden` and `Forbidden`, I think one should be removed.

XaosSintez:
@slavcodev It seems unused. I will remove it in separated PR

PR #28, opened 2020-04-15
Remove unused AccessForbidden

XaosSintez:


ashkarpetin (dismissed):

openapi/description.md
```diff
@@ -34,9 +34,6 @@ Never share your secret keys. Keep them guarded and secure.
 # Errors
 Rebilly follow's the error response format proposed in [RFC 7807](https://tools.ietf.org/html/rfc7807) also known as Problem Details for HTTP APIs.  As with our normal API responses, your client must be prepared to gracefully handle additional members of the response.
 
-## AccessForbidden
```
slavcodev, line 37: Should not be replaced with `Forbidden` ?

slavcodev:
`Forbidden` is not shown, seems path is incorrect

XaosSintez:
@slavcodev Yes, but i don't understand it yet. It is not different from `Conflict` at first glance.

slavcodev:
@XaosSintez most probably it's unused and redocly does not build it in final json.

XaosSintez:
It is used in many places. For example here:
https://github.com/Rebilly/api-definitions/blob/09e79aa7477e4e63cffd75df1effbf6fab0231c8/openapi/paths/gateway-accounts.yaml#L38

slavcodev:
This path is not included in core https://preview.redoc.ly/rebilly-core/remove-AccessForbidden/tag/Transactions

XaosSintez:
@slavcodev Yes, it looks unused in core as i see. So, no reason to add it to `description.md` then?

slavcodev:
Depends :) `description.md` is shared as far as I understand. I think we need to add `403` response in core to where is appropriate.

XaosSintez:
As i understand, it is mostly used in ACL checks (with few uses in some endpoints). ACL check are applicable to all the endpoints. So, it can be added to almost everywhere IMO.

In this case i think makes sense to add it to single endpoint here (to make `description.md` working as expected) and create separated ticket/PR to add this response to all endpoints. 

@slavcodev What do you think?

slavcodev (approved):

XaosSintez:
@slavcodev Thanks

ashkarpetin (approved):

PR #33, opened 2020-04-19
Prepare for new 3DS2 server

adamaltman:
Prepare for new 3DS2 server by enabling collection of additional browser data.


openapi/components/schemas/PaymentTokens/BrowserData.yaml
```diff
@@ -0,0 +1,63 @@
...
+      - 15
+      - 16
+      - 24
+      - 32
+      - 48
+    example: 24
+  ipAddress:
+    description: The browser's IP address (ipv4 or ipv6).
+    type: string
+    format: ipv4 or ipv6
```
slavcodev, line 26: Is this  avalid format?

2e3s: It's rendered as `string <ipv4 or ipv6>`
OpenAPI leaves this custom:
> However, format is an open value, so you can use any formats, even not those defined by the OpenAPI Specification, such as:

Although, I think it should ideally follow a regular in order to be utilized by automation tools.

adamaltman: The OpenAPI specification only supports these formats:

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#dataTypeFormat

Other formats are valid. 

However, you cannot expect tooling even to support those pre-defined formats, let alone any other format.

openapi/components/schemas/PaymentCards/PaymentCard.yaml
```diff
@@ -42,6 +42,9 @@ properties:
       contains alphanumeric values.
     type: string
     readOnly: true
+    deprecated: true
```
ashkarpetin, line 45: Why is pan fingerprint deprecated?

adamaltman: Oh. Please undo that line then. Thank you.

openapi/components/schemas/PaymentTokens/BrowserData.yaml
```diff
@@ -0,0 +1,71 @@
...
+  - colorDepth
+  - javaEnabled
+  - language
+  - screenHeight
+  - screenWidth
+  - timeZoneOffset
+  - userAgent
+properties:
+  acceptHeader:
+    description: The browser's accept header value.
```
rensftw, line 15: This is a tricky one.
I researched this extensively today, and found the following:
* Typically, the browser sets the `Accept` header
* Its value is different depending on the resource type
* **There seems to be no way to get the `Accept` value via JS in the browser**

I tested various approaches locally, using both the old `XMLHttpRequest()` and the newer `fetch()`/`Request/Response` APIs but to no avail. 
The easiest way to retrieve and store the `Accept` header value generated by the browser is server-side, although I'm not sure if this defeats the whole security aspect.

rensftw: cc: @adamaltman, @ashkarpetin, @ptrofimov

adamaltman: This should be done on the server side. The request from the browser should be sent to the server, it will include the Accept header.

adamaltman: (The API should behave like to take the "acceptHeader" from the body of the request if it is supplied, or else to take it from the actual "Accept" header. This can be done with the other headers too.)

ashkarpetin: So these fields are optional?

rensftw: One interesting thing I found was [this website](https://www.whatismybrowser.com/detect/what-http-headers-is-my-browser-sending) which pretty much mirrors the request headers a browser makes on the very first request and they have a (paid) API as well.

If the `Accept` header value is critically important, we could explore that as an option. Otherwise, we have two options:
1. Omit it.
2. Populate this field sever side.

Either way, we should probably remove it from the `BrowserData` schema

adamaltman: What if there is a client that integrates and wishes to supply the `BrowserData` from the server side?

I don't think it should be omitted.

adamaltman: @ashkarpetin no, I think only `acceptHeader` and `userAgent` are optional (should be removed from the `required` properties list). Those can be collected by the server when this API is used from the client-side.

ashkarpetin: @adamaltman ipAddress too?

From 3dsecure docs: `IP address of the browser as returned by the HTTP headers to the 3DS Requestor.`

ashkarpetin: Made acceptHeader / userAgent  optional. Will be collected from request if is not provided in payload,

adamaltman: Yes, if a server is going to use the API, they must be able to provide it. If the browser is going to provide it, we need to grab it from the HTTP headers.

rensftw: Btw, I'm all for getting the IP server side as well. 
In the current form of the library I have a dependency for this, but it's quite bad imo as it adds to the library's fingerprint and since it needs to make a request it also delays the `browserData` generation.

adamaltman: Yes, I agree you should avoid that.

ashkarpetin (dismissed):

openapi/components/schemas/PaymentTokens/BrowserData.yaml
```diff
@@ -0,0 +1,69 @@
...
+    enum:
+      - 1
+      - 4
+      - 8
+      - 15
+      - 16
+      - 24
+      - 32
+      - 48
+    example: 24
```
rensftw, line 29: What fallback values should I send if a browser API is missing?
For example, if `screen.colorDepth === undefined` or `navigator.javaEnabled === undefined`, should I send `null` in the respective field?

rensftw: cc: @adamaltman @ashkarpetin

ashkarpetin: option 1: send fallback value if value is not defined
options 2: do nothing (there is something wrong with user browser), send null, catch validation error, redirect back to merchant site

rensftw: There is great browser support for `navigator` and `screen`. 
It would be fishy if one of those APIs is missing, so I think option 2 is the better option.

image

openapi/components/schemas/PaymentTokens/BrowserData.yaml
```diff
@@ -0,0 +1,69 @@
...
+  ipAddress:
+    description: The browser's IP address (ipv4 or ipv6).
+    type: string
+    format: ipv4 or ipv6
+    example: 93.92.91.90
+  javaEnabled:
+    description: Can the browser execute Java? Value is returned from the `navigator.javaEnabled` property.
```
edmandiesamonte, line 36: Not sure if a rhetorical question is a good description here.

ashkarpetin: Thanks, addressed

edmandiesamonte (approved):

ashkarpetin (approved):

PR #35, opened 2020-04-19
Add gateway account limits

adamaltman:


openapi/components/schemas/GatewayAccountDowntimeSchedule.yaml
```diff
@@ -4,30 +4,31 @@ required:
...
       - ongoing
       - finished
   reason:
-    description: The Gateway Account downtime schedule reason
+    description: The gateway account downtime schedule reason
     readOnly: true
     type: string
     enum:
       - scheduled-maintenance
+      - limit-reached
```
XaosSintez, line 25: I see discrepancy between this `limit-reached` and `daily-limit-reached` and `monthly-limit-reached` from RFC

openapi/paths/gateway-accounts@{id}@limits@{limitId}.yaml
```diff
@@ -0,0 +1,99 @@
...
+    '401':
+      $ref: ../components/responses/Unauthorized.yaml
+    '403':
+      $ref: ../components/responses/Forbidden.yaml
+    '404':
+      $ref: ../components/responses/NotFound.yaml
+put:
+  tags:
+    - Gateway Accounts
+  summary: Update a aateway account limit
```
XaosSintez, line 40:
```suggestion
  summary: Update a gateway account limit
```

openapi/components/schemas/GatewayAccountDowntimeSchedule.yaml
adamaltman: Yes, I actually decided `limit-reached` is better. We may have an explosion. What if we need to add weekly limits? Or some limits that aren't time-based? etc..

adamaltman: (And thank you, I fixed the discrepancy in the RFC.)

XaosSintez (dismissed):

ashkarpetin (dismissed):

slavcodev (dismissed):

openapi/components/schemas/GatewayAccountLimit.yaml
```diff
@@ -0,0 +1,69 @@
...
+    enum:
+      - monitoring
+      - reached
+  startTime:
+    description: The limit's current period start time.
+    type: string
+    format: date-time
+    readOnly: true
+  endTime:
+    description: The limit's current period end time. At this time, the limit will reset.
```
svensono, line 25: Do the start- and endTime mean the period in which the limit is active? For example: 
**startTime**: 1-1-2020 00:00:00 
**endTime**: 31-12-2020 23:59:59
**frequency**: daily

Or are they related to the times of the frequency?
**startTime**: 1-1-2020 09:00:00
**endTime**: 1-1-2020 18:00:00
**frequency**: daily

XaosSintez: It is current period. If it is maontly period, it will be 00:00 1st day of this month as a startTime and 23:59:59 of last day of this month as endTime

adamaltman:
@VladimirBerdnik see the #891db8c commit with a change related to discussions about improving the structure to prevent the rabbit hole. Since it eliminates one API operation, I also think this reduces the scope. 🛠️

VladimirBerdnik:
> @VladimirBerdnik see the [commit](https://github.com/Rebilly/api-definitions/pull/35/commits/891db8c110cdb7060d968614fb809a8e30244f79) with a change related to discussions about improving the structure to prevent the rabbit hole. Since it eliminates one API operation, I also think this reduces the scope.

@adamaltman , should we change these fields to read only to not to give API consumers make conflicts of identifiers with these fields? 
image

adamaltman:
Yes @VladimirBerdnik, definitely.

VladimirBerdnik:
@adamaltman , @slavcodev suggested to use IDs in reverse order like these:

`daily-money-limit`
`monthly-count-limit`

They looks more user friendly to read and pronounce, will change if there is no any objections

adamaltman:
I don't have any objection.

edmandiesamonte (approved):

ashkarpetin (approved):

iamdevice (approved):

PR #37, opened 2020-04-20
Add round robin pick gateway account strategy

adamaltman:


openapi/components/schemas/Rules/GatewayAccountPickInstruction.yaml
```diff
@@ -5,6 +5,15 @@ discriminator:
     gateway-account-weights: ./PickInstructions/gateway-account-weights.yaml
     gateway-acquirer-weights: ./PickInstructions/gateway-acquirer-weights.yaml
 properties:
+  strategy:
+    type: string
+    enum:
+      - weighted-random
```
slavcodev, line 11: Should we add `defaults` keywords?

adamaltman: Possibly. I guess that would be easiest.

PR #47, opened 2020-04-23
Add StripeMPI

vova07:
Add `StripeMPI` definition.

### DIFF
--- a/openapi/components/schemas/Mpi/StripeMpis/StripeMpi.yaml
+++ b/openapi/components/schemas/Mpi/StripeMpis/StripeMpi.yaml
@@ -1,0 +1,8 @@
+description: Stripe Integrated
+allOf:
+  - $ref: ./StripeMpis.yaml
+  - type: object
+    properties:
+      requestThreeDSecureForAny:
+        type: boolean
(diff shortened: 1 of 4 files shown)


openapi/components/schemas/Mpi/StripeMpis/StripeMpi.yaml
```diff
@@ -0,0 +1,8 @@
+description: Stripe Integrated
+allOf:
+  - $ref: ./StripeMpis.yaml
+  - type: object
+    properties:
+      requestThreeDSecureForAny:
```
ashkarpetin, line 6: enforceThreeDSecure?

ashkarpetin: What does mean `request for any`?

vova07: I thought about that name, but the idea is that it will enforce it, only where it will be possible.
But maybe it's better to rename it.

slavcodev (approved):

ashkarpetin (approved):

iamdevice (approved):

PR #58, opened 2020-05-05
Add new timeline transaction gateway connection messages

destebang:
Added two new timeline transaction messages regarding gateway connection error and gateway timeout.

### DIFF
--- a/openapi/components/schemas/Timelines/TransactionTimeline.yaml
+++ b/openapi/components/schemas/Timelines/TransactionTimeline.yaml
@@ -18,6 +18,8 @@ properties:
       - dispute-lost
       - dispute-responded
       - dispute-won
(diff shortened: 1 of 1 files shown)


openapi/components/schemas/Timelines/TransactionTimeline.yaml
```diff
@@ -36,6 +36,8 @@ properties:
       - transaction-capture-delayed
       - transaction-captured
       - transaction-waiting-gateway
+      - gateway-connection-timed-out
+      - gateway-connection-failed
```
ashkarpetin, line 40: Please sort alphabetically

destebang: I was just doing that, you're fast 😄

PR #80, opened 2020-05-26
Add JS examples from old SDK docs

mathewsonke:
Adds all of the JS examples which were present on https://rebilly.github.io/rebilly-js-sdk/

### DIFF
--- a/openapi/paths/customers@{id}.yaml
+++ b/openapi/paths/customers@{id}.yaml
@@ -32,6 +32,9 @@ get:
     - lang: PHP
       source:
         $ref: '../code_samples/PHP/customers@{id}/get.php'
+    - lang: JavaScript
+      source:
+        $ref: '../code_samples/JavaScript/customers@{id}/get.js'
 put:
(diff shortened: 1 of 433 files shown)


adamaltman (approved):
I didn't review all of this (it's giant). But I think it's fine to try.

openapi/paths/customers@{id}.yaml
```diff
@@ -63,6 +66,9 @@ put:
     - lang: PHP
       source:
         $ref: '../code_samples/PHP/customers@{id}/put.php'
+    - lang: JavaScript
+      source:
+        $ref: '../code_samples/JavaScript/customers@{id}/put.js'
 delete:
```
ZengJunyong, line 72: I noticed not every method(e.g. merge and delete a customer) has its PHP/Javascript examples, is this intentional? or we ignore it for now?

ZengJunyong (approved):
I picked about 10 examples to run them, all good.

PR #83, opened 2020-05-28
Add support for multiple fileIds in Kyc document

subenson:


openapi/components/schemas/Common/CommonKycDocument.yaml
```diff
@@ -1,6 +1,6 @@
 type: object
 required:
-  - fileId
+  - fileIds
```
adamaltman, line 3: Requiring it is a breaking change.

openapi/components/schemas/Common/CommonKycDocument.yaml
```diff
@@ -13,16 +13,25 @@ properties:
...
     enum:
       - identity-proof
       - address-proof
+  documentSubtype:
+     description: Document subtype submitted for validation.
+     type: string
+     enum:
+      - passport
+      - id-card
+      - drivers-license
```
adamaltman, line 34:
```suggestion
      - driver-license
```

openapi/components/schemas/Common/CommonKycDocument.yaml
```diff
@@ -13,16 +13,25 @@ properties:
     description: The сustomer's ID
     allOf:
       - $ref: ../ResourceId.yaml
-  fileId:
-    description: Linked file object id
```
adamaltman, line 17: I don't think we can remove it. We can mark it as deprecated. We need to be able to support a transition window...

openapi/components/schemas/Common/CommonKycDocument.yaml
svensono: You're right. I forgot to take BC into account in this repo. This is now  added.

agualis (approved):

ZengJunyong (approved):

PR #115, opened 2020-06-19
Add the x-multiline flag for sftpPrivateKey

ZengJunyong:
The `x-multiline` flag is used for Gateway properties that accept an RSA key


rensftw (approved):
LGTM!

iamdevice (approved):

PR #117, opened 2020-06-25
Reorder paths

adamaltman:
Reorders paths for display purposes in the API docs.

### DIFF
--- a/openapi/core.yaml
+++ b/openapi/core.yaml
@@ -210,14 +210,14 @@ x-tagGroups:
       - Customers Timeline
   - name: Payment Instruments
     tags:
+      - Payment Tokens
       - Bank Accounts
       - Payment Cards
-      - Payment Tokens
       - PayPal Accounts
   - name: Payments
     tags:
-      - 3D Secure
       - Transactions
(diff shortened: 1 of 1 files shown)


iamdevice (approved):

openapi/core.yaml
```diff
@@ -362,8 +362,8 @@ paths:
     $ref: ./paths/payment-cards.yaml
   '/payment-cards/{id}':
     $ref: './paths/payment-cards@{id}.yaml'
-  '/payment-cards/{id}/authorization':
-    $ref: './paths/payment-cards@{id}@authorization.yaml'
+  # '/payment-cards/{id}/authorization':
```
XaosSintez, line 365: If we don't need this endpoint here, may be we should just remove it instead of commenting it out?

adamaltman: I'm commenting it to prevent the display. But I want to leave it intentionally to remind us to give notice to people still using this API.

XaosSintez (approved):

subenson (approved):

VladimirBerdnik (approved):

PR #130, opened 2020-07-02
Add Plaid settings for Stripe gateway account

slavcodev:
image


adamaltman (dismissed):

vova07:
I'm pretty agree with this proposal, but I have some doubts even I don't have an alternative solutions.

My concern is related to the fact that `Plaid` MUST be used "de fatco" for all `ACH` transactions that goes through `Stripe` gateway.

In that way, from the proposal we get a case when we can select the `payment method = ACH` and save the gateway without any errors, but in fact we should send an alert and require the merchant to fill the `Plaid` credentials.

I propose to leave only `Plaid` credentials, and require them based on the selected payment method, without any needs in specifying the `Use Plaid` option at all.

What do you think?

slavcodev:
I'm fine to remove `usePlaid` and add it when we will add other ways to process ACH payments

vova07 (dismissed):

openapi/components/schemas/GatewayAccountConfig/Stripe.yaml
```diff
@@ -12,6 +12,23 @@ allOf:
...
+                type: string
+                description: Plaid public key.
+              secret:
+                type: string
+                description: Plaid secret key.
+                format: password
+            required:
+              - clientId
+              - publicKey
+              - secret
```
slavcodev, line 31:
```suggestion
          clientId:
            type: string
            description: Plaid client ID is required for the `bank-account` payment method.
          publicKey:
            type: string
            description: Plaid public key is required for the `bank-account` payment method.
          secret:
             type: string
             description: Plaid secret key is required for the `bank-account` payment method.
             format: password
```

slavcodev: I think backend can't work with nested credentials keys.

vova07: At least it require additional adjustments in multiple places to handle that.

So I agree with new proposal.

slavcodev: It also will break Recomm I guess, I've checked a dozen of other definitions, we seems do not support objects for credentials, only scalars (string, boolean)

ashkarpetin (approved):

vova07 (approved):

iamdevice (approved):

PR #140, opened 2020-07-10
Fix Stripe config definition

slavcodev:
The returnUrl is not a credential, it is a setting

### DIFF
--- a/openapi/components/schemas/GatewayAccountConfig/Stripe.yaml
+++ b/openapi/components/schemas/GatewayAccountConfig/Stripe.yaml
@@ -8,10 +8,6 @@ allOf:
       credentials:
         type: object
         properties:
-          redirectUrl:
-            type: string
-            description: Stripe Gateway redirect url
-            format: url
           plaidClientId:
             type: string
             description: Plaid client ID is required for the `bank-account` payment method.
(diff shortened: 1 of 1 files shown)


vova07 (dismissed):

ashkarpetin (dismissed):

vova07 (approved):
It will require F-E adjustments.

iamdevice (approved):

openapi/components/schemas/GatewayAccountConfig/Stripe.yaml
```diff
@@ -22,12 +18,16 @@ allOf:
...
-          - redirectUrl
       settings:
         type: object
         description: Stripe settings object
+        required:
+          - redirectUrl
         properties:
+          redirectUrl:
+            type: string
+            description: The URL where the user will be redirected after authorizing the account on Stripe.
```
adamaltman, line 29: Why is this a setting? Why isn't this following the same redirectUrl behavior we have everywhere?

slavcodev: This how it works on backend. We store this in settings. I believe changing the API should be separate project.

adamaltman: It seems this should be read only then. Stripe only allows our preconfigured URLs for oauth2.

slavcodev: No, it doesn't. [Stripe accept redirect URL dinamically](https://stripe.com/docs/connect/oauth-reference#redirect-uri)

PR #137, opened 2020-07-07
Add digital wallet payment method #230

VladimirBerdnik:

PR #148, opened 2020-07-20
Proof of address matches

subenson:


openapi/components/schemas/KycDocument/ProofOfIdentity.yaml
```diff
@@ -0,0 +1,82 @@
...
+                example: John
+              lastName:
+                description: 'The customer last name if it was matched, null otherwise'
+                type: string
+                example: Doe
+              dateOfBirth:
+                description: 'The date of birth found on the document, null if not found'
+                allOf:
+                  - $ref: ../ServerTimestamp.yaml
+              expiryDate:
```
ashkarpetin, line 59: That still not correct, we don't use British english, should be expiration date. e.g.: https://github.com/Rebilly/api-definitions/pull/137#discussion_r452809063

svensono: I'm in favor of changing this in the API, however XY uses the KYC API. So I would like to pick this is up separately, outside of the scope of this RFC.

ashkarpetin (approved):

ptrofimov (approved):

PR #153, opened 2020-07-27
Add digital wallet validation endpoint

VladimirBerdnik:
Add endpoint to validate an Apple Pay session


openapi/components/schemas/PaymentTokens/DigitalWalletValidation/ApplePayValidation.yaml
```diff
@@ -0,0 +1,33 @@
...
+    readOnly: true
+  merchantSessionIdentifier:
+    type: string
+    readOnly: true
+  nonce:
+    type: string
+    readOnly: true
+  signature:
+    type: string
+    readOnly: true
```
colinbird, line 33: I assume all of these readOnly properties are what is delivered in the payload from Apple Pay ?

VladimirBerdnik: @colinbird , yes

VladimirBerdnik: I wonder if it is better to wrap such properties to simple `validationRequest` and `validationResponse` objects as they both are used by the SDK

colinbird: Yes, I think that may be better if this may be reused for other wallets in the future.  Although perhaps it's unlikely that will happen

VladimirBerdnik: got it, thanks, will change

colinbird (dismissed):

subenson (approved):

iamdevice (approved):

PR #156, opened 2020-07-30
Define `x-webhooks` in API definitions

mathewsonke:
- Adds webhook definitions for each eventType
- Updates the GlobalWebhookEventTypes list
- Adds Embeds and Links for some resources
- Defines request bodies for webhooks
- Sorts the EventType list

Our webhooks can use any HTTP verb and they will provide the same information. The webhooks spec however does not appear to allow us to define webhooks which are verb-agnostic. For that reason I have defined all the webhooks as `post`.


mathewsonke:
The preview build check was (and is) failing, but for some reason it is not being counted as a check anymore, and I can't trigger a preview build manually either. I'm still investigating.

openapi/storefront.yaml
```diff
@@ -43,7 +43,7 @@ x-tagGroups:
       - Authentication
       - Email verification
       - Password reset
-      - KYC Document
+      - KYC Documents
```
adamaltman, line 46:
```suggestion
      - KYC documents
```

openapi/webhooks/customer-created.yaml
```diff
@@ -0,0 +1,9 @@
+summary: Customer created
+operationId: customer-created
+tags:
+  - Customers
+requestBody:
+  $ref: ../components/schema/requestBodies/Webhooks/Customer.yaml
+responses:
+  '200':
```
adamaltman, line 8: We accept any 200-series response. I think you can use `2XX` with uppercase `X`: 

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#patterned-fields-1

adamaltman: Same for all of the other webhook responses... (I won't leave the same comment everywhere)

mathewsonke:
I believe the build failure is unavoidable until this PR in Redoc is merged: https://github.com/Redocly/redoc/pull/1304

Right now, `openapi-cli` won't properly bundle the external files referenced from `x-webhooks` which results in a 404 when trying to preview the docs. 

@adamaltman to avoid build problems, should I hold off merging this PR until `x-webhooks` support is deployed by Redoc?

adamaltman:
That shouldn’t be tied to build failure.

mathewsonke:
Update- I reached out to Roman a couple days ago to help me investigate the build issues and the bug Adam reported, but he's got a lot of work right now and hasn't had a chance to get to it yet.

adamaltman:
@mathewsonke the issue is openapi-cli doesn't bundle the specification extensions. I opened this bug: https://github.com/Redocly/openapi-cli/issues/174 (feel free to contribute there if you want to fix it ASAP... if not, it's okay. The solution may be in adjusting here (somewhere): https://github.com/Redocly/openapi-cli/blob/master/src/typings/openapi.ts

adamaltman (approved):

PR #161, opened 2020-08-05
Add Storefront API link to README

slavcodev:


README.md
```diff
@@ -7,6 +7,7 @@ TBD: Add Redocly validation status badge.
 - [Core API docs](https://api-reference.rebilly.com/)
 - [Users API docs](https://user-api-docs.rebilly.com/)
 - [Reports API docs](https://reports-api-docs.rebilly.com/)
+- [Storefront API docs](https://storefront-api-docs.rebilly.com/)
```
ashkarpetin, line 10: It requires authentication

slavcodev: Right. What is your concern?

ashkarpetin: It's a public repo, all links should be public

slavcodev: Why all links should be public? As you said the repository is public, the API defnitions is open-sourced, so existence of such API is not a big secret. I have no idea why the docs is closed for auth users only, probably to limit integrations, but I don't see a reason to not have a links for those who have an access to it.

ashkarpetin (approved):

VladimirBerdnik (approved):

PR #166, opened 2020-08-05
Add Plaid Account Token to Tokens API

slavcodev:
image


slavcodev:
php please review

openapi/components/schemas/PaymentTokens/PaymentToken.yaml
```diff
@@ -2,6 +2,7 @@ type: object
 title: Payment Token
 allOf:
   - $ref: ../Common/CommonPaymentToken.yaml
+  - $ref: ./PaymentTokenDeprecatedProperties.yaml
```
adamaltman, line 5: I would remove all of these... I think this is a bad addition.

adamaltman (dismissed):
I don't like adding deprecated properties as a separate file, and I believe we intentionally removed it from the documentation before.

slavcodev:
We didn't remove them from docs before. Now it is commen fields, I've removed it in this PR 
https://github.com/Rebilly/api-definitions/pull/166/files#diff-70113a5a1e0f61a2a6e506a45c42bf01L17

I segregate it in new file to not add to new token types like we did for `DififtalWalletToken`.
Let me know if we need to remove it from all tokens, I'm 👍  for that

openapi/components/schemas/PaymentTokens/PaymentToken.yaml
slavcodev: It's not an additional, just to make clear. See `CommonPaymentToken`
 https://github.com/Rebilly/api-definitions/pull/166/files#diff-70113a5a1e0f61a2a6e506a45c42bf01L17

adamaltman:
Let's remove it from all tokens (the backend will continue to support it). 

It's been marked as deprecated for a while, and nobody should get an idea that they should try it now.

ashkarpetin (dismissed):

subenson (dismissed):

slavcodev:
php please review

ashkarpetin (approved):

wmdmgv (approved):

PR #177, opened 2020-08-19
Add manual taxes and shipping to Invoice

XaosSintez:


openapi/components/schemas/Common/CommonInvoice.yaml
```diff
@@ -84,8 +101,9 @@ properties:
       $ref: ../Invoices/InvoiceItem.yaml
   taxes:
     type: array
-    description: Taxes applied to this invoice
-    readOnly: true
+    description: >-
+      Taxes applied to this invoice.
+      To apply manual taxes, set "taxCalculator" to "manual".
```
2e3s, line 106: Instead of this and 
>  If "manual", "taxes" field should be presented.

I suggest to use a discriminator on `taxCalculator` value.

2e3s: Same for `shippingCalculator`.

svensono: I'll create a InvoiceRequest.yaml with discriminator, since the taxes are included in the result of both taxCalculators

2e3s: I've created another PR below with the discriminators.

ashkarpetin (approved):

PR #190, opened 2020-08-27
Add PostPayout operation

ashkarpetin:


openapi/components/schemas/Transactions/TransactionCreditRequest.yaml
```diff
@@ -0,0 +1,88 @@
...
+  - websiteId
+  - customerId
+  - currency
+  - amount
+properties:
+  websiteId:
+    description: The website identifier string.
+    allOf:
+      - $ref: ../ResourceId.yaml
+  customerId:
```
slavcodev, line 12: Who is the customer of payout?

openapi/components/requestBodies/PayoutRequest.yaml
```diff
@@ -0,0 +1,6 @@
+content:
+  application/json:
+    schema:
+      $ref: ../schemas/Transactions/PayoutRequest.yaml
```
slavcodev, line 4: What difference between this object and Trnsaction? I'm confusing why we post one object and receive another

ashkarpetin: No deprecated properties,  no `type` ( for /payouts transaction it's credit), different descriptions.

slavcodev: Does it make sense to use old response schema then?

ashkarpetin: I think so, anyway i can't use original schema, because descriptions are different

slavcodev: I'm confusing, can you elaborate why deprecated properties make no sense for request, but do make for response? The same question about `type`.

ashkarpetin: > Can you elaborate why deprecated properties make no sense for request,

I did not say that.

If you want I can return deprecated properties and `type` back

slavcodev: > I did not say that.

You didn't say, but you decided that new schema is necessary for reuqest to remove deprecated fields.

> If you want I can return deprecated properties and type back

I did not mean this. I mean if we going to strip deprecated properties from new API (which I think makes sense), we also should strip them from response. Because in **Evolution API** the maintenance of derecated stuff in responses is more complicated. You can monitor API log for requests and remove deprecated proeprties if no one use it, but you never know how clients integrate your API and how they handle responses.

ashkarpetin: That will require additional transformers on backend side specifically for this endpoint, is it worth investing time implementing it without no benefit?

slavcodev: I believe it's worth to clear responses and requests from deprecated stuff in new API.

slavcodev: Closing this, I've misunderstood the result of this new operation.
Now I see it returns the created "credit" transaction, i.e. the created resources is located by `GetTransaction` URL.

openapi/components/schemas/Transactions/PayoutRequest.yaml
```diff
@@ -0,0 +1,88 @@
+type: object
+required:
+  - websiteId
+  - customerId
+  - currency
+  - amount
+properties:
```
slavcodev, line 7: What is difference between this and `TransactionRequest.yaml`? Can we re-use, or use some knidnof composition in order to not duplicate the schemas?

svensono: Transaction has a `type` field whereas Payout has not. I moved all the shared fields (all payouts fields) to CommonTransactionRequest.

slavcodev (approved):

iamdevice (approved):

PR #218, opened 2020-09-25
Add polymorphic taxes and shipping to Invoice

2e3s:
With `manual` and `rebilly` based calculators.

This is an improvement over #177 because a few of certain aspects of invoices obtain a more complex behaviour and the given way works better with OAS and automated tools.
The older fields will be kept for BC since they don't intersect the new ones.

[Preview](https://preview.redoc.ly/rebilly-core/invoice-poly-taxes-shipping/tag/Invoices#operation/PostInvoice)


openapi/components/schemas/Common/CommonInvoice.yaml
```diff
@@ -55,11 +55,12 @@ properties:
     type: number
     format: double
     readOnly: true
```
ashkarpetin, line 57: I wonder about this, you removed shippingAmount but left this

openapi/components/schemas/Common/CommonInvoice.yaml
```diff
@@ -55,11 +55,12 @@ properties:
...
-  shippingAmount:
```
ashkarpetin, line 58: You removed `shippingAmount` but left `taxAmount`. I wonder if it's intentional?

openapi/components/schemas/Invoices/TaxTypes/rebilly-taxjar.yaml
```diff
@@ -0,0 +1,13 @@
+description: Rebilly-based shipping amount calculation with shipping zones.
```
ashkarpetin, line 1: Seems incorrect description

openapi/components/schemas/Invoices/InvoiceShipping.yaml
```diff
@@ -0,0 +1,16 @@
+description: Invoice shipping
+type: object
+discriminator:
+  propertyName: calculator
+  mapping:
+    manual: ./ShippingTypes/manual.yaml
+    rebilly: ./ShippingTypes/rebilly.yaml
+properties:
+  calculator:
```
slavcodev, line 9: `calculator: manual` does not sound natural, may be we can use another term, e.g. `method` or `type`?

openapi/components/schemas/Common/CommonInvoice.yaml
slavcodev: Yes. It's migrated to `shipping.amount`. We can add `deprecated` flag to these, but I think it's better to remove them from docs to make sure new integrators will not use them mistakenly.

slavcodev (dismissed):

openapi/components/schemas/Common/CommonInvoice.yaml
2e3s: Yes, I've removed that too. Deprecation is interesting when generating SDK clients, but we don't do that.

ashkarpetin (dismissed):

slavcodev (dismissed):

openapi/components/schemas/Common/CommonInvoice.yaml
```diff
@@ -44,16 +44,12 @@ properties:
...
-    description: The invoice's shipping amount.
-    type: number
-    format: double
-    readOnly: true
+  shipping:
+    allOf:
+      - $ref: ../Invoices/InvoiceShipping.yaml
+  tax:
+    allOf:
```
adamaltman, line 51: Why allOf and not just $ref?

subenson: There is no valid reason for this. I changed it

ashkarpetin (approved):

justnero (approved):

PR #220, opened 2020-09-29
Support iDEAL for IpayOptions

colinbird:


openapi/components/schemas/GatewayAccountConfig/IpayOptions.yaml
```diff
@@ -39,3 +39,10 @@ allOf:
             enum:
               - SOAP
               - TxHandler
+              - SecureHosted
+          cardType:
```
colinbird, line 43: IpayOptions uses this terminology, so I thought it best to use that here even though it's not really a card type.

openapi/components/schemas/GatewayAccountConfig/IpayOptions.yaml
```diff
@@ -39,3 +39,10 @@ allOf:
...
+            type: string
+            description: Manually set the cardType for iDEAL.
```
colinbird, line 45:
```suggestion
            description: Manually set the card_type for iDEAL.
```

2e3s (approved):

XaosSintez (approved):

PR #233, opened 2020-10-08
Simplify Ready to Pay adjust action

justnero:


openapi/components/schemas/Rules/Actions/adjust-ready-to-pay.yaml
```diff
@@ -1,8 +1,14 @@
...
   - $ref: ../RuleAction.yaml
   - type: object
-    description: Add risk score
+    description: Adjust ready to pay
     properties:
+      pushInstrumentsUp:
+        type: boolean
+        default: false
+        description: |
+          If set as `true`, the payment methods that a customer has a saved payment instrument for will be displayed at
```
adamaltman, line 10: I think it should be "active" payment instruments.
```suggestion
          If set as `true`, the payment methods that a customer has active payment instruments for will be displayed at
```

2e3s (dismissed):

2e3s:
But sometimes merchants need to refine a priority for payment methods more than that.

justnero:
This will preserve the ordering between the pushed methods. If more fine tuning is required merchant could use manual adjustments using separate rules with conditions

openapi/components/schemas/Rules/Actions/adjust-ready-to-pay.yaml
```diff
@@ -1,8 +1,14 @@
...
   - $ref: ../RuleAction.yaml
   - type: object
-    description: Add risk score
+    description: Adjust ready to pay
     properties:
+      pushInstrumentsUp:
+        type: boolean
+        default: false
+        description: |
+          If set as `true`, the payment methods that a customer has active payment instruments for will be displayed at
```
adamaltman, line 10:
```suggestion
          If set as `true`, the payment methods associated to the customer's active payment instruments will be displayed at
```

openapi/components/schemas/Rules/Actions/adjust-ready-to-pay.yaml
```diff
@@ -1,8 +1,14 @@
 allOf:
   - $ref: ../RuleAction.yaml
   - type: object
-    description: Add risk score
+    description: Adjust ready to pay
     properties:
+      pushInstrumentsUp:
```
svensono, line 6: Maybe it's because I'm not very familiair with it yet, but the key pushInstrumentsUp seems a bit vague. What do you think about something like prioritizeActivePaymentInstrument?

justnero: I don't like the word `prioritize` in this context. Generally we push the weights up preserving their order while prioritize would mean we are reordering active payment instrument methods no matter what weight it was.

adamaltman: Then perhaps the feature is still off? I think `prioritize` is more like what I described in the meeting the other day.

adamaltman: The concept of weights is actually incorrect and inconsistent with how we use weights throughout our app. Weights are used for weighted random selections. We want ordered sequences, not weights.

adamaltman (changes requested):
I actually think the smarter approach would be to redo the action entirely. The conceptual logic is wrong, and will just come to bite us in the end.

Thinking of it in terms of the simplicity cycle, by leaving all of that weights, allow/deny stuff in there we have:

a) more complexity
b) less goodness

adamaltman:
This is looking better. What is the (default) behavior if no rule action is set/matched?

justnero:
In such case no methods are presented in the ready to pay response.

justnero:
If no actions matched we would expose all methods available assuming Gateway Account filters. To invert this behavior you can place an all-matching rule at the end of the list with an empty `paymentMethods` array.

adamaltman:
This sounds good to me. This information should be written somewhere in the documentation (not just in comments here).

adamaltman (dismissed):

adamaltman (approved):

adamaltman:
Please adjust the PR title.

ashkarpetin (approved):

iamdevice (approved):

PR #239, opened 2020-10-19
Add links to storefront resources

2e3s:
This corresponds to [HATEOAS](https://en.wikipedia.org/wiki/HATEOAS) in a certain degree. There could be more links to control the entities, the given ones are the most basic which also exist in the main API.


openapi/components/schemas/Storefront/StorefrontTransaction.yaml
```diff
@@ -5,3 +5,13 @@ allOf:
         description: The URL to redirect the end-customer when transaction `status` is `waiting-approval`.
         type: string
         format: uri
+      _links:
```
ashkarpetin, line 8: approvalUrl seems duplicated ApprovalUrlLink, storefront doesn't use _links, so I'm not sure about this change

justnero: Agreed. Storefront doesn't use neither `_links`, not `_embedded` concept, why do we need it?

2e3s: This is unrelated to `_embedded`.

How to pass links to Storefront then? To cancel etc. Add an entity field `cancelUrl`?

2e3s: I would think that `_links` is a more graceful concept.

justnero: Ok, then you should remove `approvalUrl` entity field. There should not be any duplicates. Also we should be consistent with our API – if we add `_links` here we should include it everywhere else.

2e3s: I could find control URLs only for Transaction entity. So other entities would only need to get self link.
As I see, there is `_links` also in Website. Because the common entity is used.

justnero (dismissed):

openapi/components/schemas/CheckoutForm/StorefrontCheckoutForm.yaml
```diff
@@ -6,3 +6,11 @@ allOf:
         description: The session's token used for authentication
         type: string
         readOnly: true
+      _links:
+        type: array
+        description: The links related to resource
```
vykhrystiuk, line 11:
```suggestion
        description: The links related to resource.
```

I think we need a period there (and other places)

https://rebilly.slack.com/archives/C7U0AHYJ0/p1601565453042200

vykhrystiuk (dismissed):

justnero (dismissed):

iamdevice (dismissed):

slavcodev (dismissed):

openapi/components/schemas/CheckoutForm/StorefrontCheckoutForm.yaml
2e3s:
```suggestion
        description: The links related to resource.
```

slavcodev (approved):

vykhrystiuk (approved):

iamdevice (approved):

justnero (approved):

PR #253, opened 2020-10-27
Add parameter for SubscriptionChangePlan bool 'keepTrial'

lexerom:
https://api-reference.rebilly.com/#operation/PostSubscriptionPlanChange - additional parameter bool "keepTrial". Allows to keep trial on plan change with 'retain' renewal policy.

### DIFF
--- a/openapi/components/schemas/SubscriptionChange.yaml
+++ b/openapi/components/schemas/SubscriptionChange.yaml
@@ -47,7 +47,7 @@ properties:
     default: false
   keepTrial:
     description: >-
-      If set to true and the old plan has an active trial, that trial will be used for the new plan. Works with 'retain'
+      If set to true and the subscription order has an active trial, it will use that trial further. Works with 'retain'
       renewalPolicy only.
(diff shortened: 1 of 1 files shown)


openapi/components/schemas/SubscriptionChange.yaml
```diff
@@ -45,3 +45,9 @@ properties:
       to preview the changes that would be made to a subscription.
     type: boolean
     default: false
+  keepTrial:
+    description: >-
+      If set to true and the old plan has an active trial, that trial will be used for the new plan. Works with 'retain'
```
2e3s, line 50: This API should have no effect on plans.
```suggestion
      If set to true and the subscription order has an active trial, it will use that trial further. Works with 'retain'
```

VladimirBerdnik (approved):

lexerom:
php please review
js please review

ashkarpetin (approved):

vykhrystiuk (approved):

PR #254, opened 2020-10-28
Storefront Transaction processing API

justnero:


2e3s:
Would this be used instead of PATCH transaction with its subsequent continuation?

justnero:
We may transform `StorefrontPatchTransaction` to be a part of this API in future, but for now it's out of scope.

openapi/components/requestBodies/storefront/PostTransactionProcessing.yaml
```diff
@@ -0,0 +1,21 @@
...
+      properties:
+        action:
+          type: string
+          description: The action customer performs to change current transaction's processing state.
+          enum:
+            - submit
+            - cancel
+        params:
+          type: object
+          description: Contains all the form values required by the current processing state.
```
2e3s, line 17: What would it have? It looks like one API endpoint that rules them all.

justnero: It would have values for the `fields` (at least required ones) sent in `StorefrontGetTransactionProcessing` response with `type: form`. In other words this wraps the `FormBody` from the frontend.

2e3s: I think all the fields should be described in a contract. Based on what those fields will be routed? Smartly defined by the provided fields?

2e3s: Currently we distinguish what to do by the hook URL where the customer is redirected.
This endpoint looks like the one to rule them all.

justnero: Describing all the fields for each and every gateway integration would lead us to a great amount of hardcoded integrations in API definitions. This endpoint only collects the feedback from the `form` and `confirm` processing states.

2e3s: > a great amount of hardcoded integrations in API definitions

I prefer the word "documented" in lieu of "hardcoded".

> This endpoint only collects the feedback

It receives form data from the forms as I see.

2e3s: And I still don't understand how the endpoint would interpret the incoming form data.

justnero: It receives the form data based on the dynamic form configuration from `StorefrontGetTransactionProcessing` and interprets it based on the same assumptions used to initially generate that configuration. "Documenting" each and every possible form configuration would be almost the same as implementing an alternative you've proposed to the RFC.

openapi/components/requestBodies/storefront/PostTransactionProcessing.yaml
```diff
@@ -0,0 +1,21 @@
...
+      type: object
+      required:
+        - action
+        - browserData
+      properties:
+        action:
+          type: string
+          description: The action customer performs to change current transaction's processing state.
+          enum:
+            - submit
```
adamaltman, line 13: what does `submit` mean?

openapi/components/schemas/Storefront/TransactionProcessing.yaml
```diff
@@ -0,0 +1,15 @@
+type: object
+properties:
+  type:
+    type: string
+    description: The type of the required action.
+    enum:
```
adamaltman, line 6: This mixes nouns and verbs. `confirmation` would be the noun. I don't know quite what `async` is in this context.

openapi/components/requestBodies/storefront/PostTransactionProcessing.yaml
justnero: The actual form submit by a customer.

openapi/components/schemas/Storefront/TransactionProcessing.yaml
justnero: >I don't know quite what async is in this context.

I've described it here
https://github.com/Rebilly/api-definitions/pull/254/files#diff-8d51afea60df179814395c04012e9e855373a1aece7d7ef086b55b5eb099e89fR26-R27

openapi/components/schemas/Storefront/TransactionProcessing/ProcessingRedirect.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - get
+      - post
+  params:
+    type: object
+    description: |
+      The extra payload for the redirect.
+
+      If `method` is `get` should be treated as a query string params.
+
+      If `method` is `post` should be treated as a POST body.
```
slavcodev, line 24: I don't think "redirect" can use other method than "GET"

slavcodev: The form which requries "POST" is not a redirect is a submit

justnero: That doesn't really mean we need to use `window.location.href = ...`. Redirect type means we need redirect customer to some location. Generic means to do this would be to autosubmit the form with a specified method.

openapi/components/schemas/Storefront/TransactionProcessing/ProcessingForm.yaml
```diff
@@ -0,0 +1,14 @@
+type: object
+title: Form
+properties:
+  type:
+    enum:
+      - form
+  fields:
```
slavcodev, line 7: What about buttons? We have submit (okay can be hardcoded on HPF), on some pages we also have "cancel" and maybe other

openapi/components/schemas/Storefront/TransactionProcessing/ProcessingConfirmation.yaml
```diff
@@ -0,0 +1,8 @@
+type: object
+title: Confirmation
+properties:
+  type:
+    enum:
+      - confirmation
```
slavcodev, line 6: What is difference between `form`? I guess the confirmation means "PostTransactionProcessing" with one value "true" or "confirmed", or whatever that backend can treat as confirmation.

openapi/components/schemas/Storefront/TransactionProcessing/ProcessingAsync.yaml
```diff
@@ -0,0 +1,11 @@
+type: object
+title: Async
```
slavcodev, line 2: This does not seem to be a type. More like status of processing.
It also `async` is not a noun and odes not fit to other types, like Adam mentioned.

openapi/components/schemas/Storefront/TransactionProcessing/FieldTypes/ProcessingFormSelectField.yaml
```diff
@@ -0,0 +1,36 @@
+type: object
+title: Select
+properties:
+  type:
+    enum:
+      - select
+  appearance:
```
slavcodev, line 7: Why not just create more types?

openapi/components/schemas/Storefront/TransactionProcessing/FieldTypes/ProcessingFormPanField.yaml
```diff
@@ -0,0 +1,15 @@
+type: object
+title: PAN
+properties:
+  type:
+    enum:
+      - pan
+  supportsLuhn:
```
slavcodev, line 7: Is there any `pan` that do not support Luhn?

openapi/components/schemas/Storefront/TransactionProcessing/FieldTypes/ProcessingFormNumberField.yaml
```diff
@@ -0,0 +1,22 @@
+type: object
```
slavcodev, line 1: It seems this pretty much re-inventing Json-Schema, may be better to adopt it as  standard? Or it also represent the `HTML tag`, so may be use those terms as standard? E.g. `multipleOf` is called "step" in HTML.

openapi/components/schemas/Storefront/TransactionProcessing/FieldTypes/ProcessingFormPanField.yaml
ashkarpetin: China UnionPay

openapi/components/schemas/Storefront/TransactionProcessing/ProcessingRedirect.yaml
slavcodev: Ahh it is designed for autosubmit form. Okay, the assumption that this form is list of hidden fields that can be set in `params` may work. Or might not, for every gateway.

openapi/components/schemas/Storefront/TransactionProcessing/ProcessingForm.yaml
justnero: You can take a look at `information` field that contains buttons. There is a `cancel` button and it should be `null` if we don't need it.

openapi/components/schemas/Storefront/TransactionProcessing/ProcessingConfirmation.yaml
justnero: No fields, we can combine them I think.

openapi/components/schemas/Storefront/TransactionProcessing/FieldTypes/ProcessingFormSelectField.yaml
justnero: This way we would allow frontend to decide what will fit better based on the options quantity, layout, screen size and so on. In other words it is better to show two radio/checkboxes instead of a select dropdown for two elements, but 10 radio/checkboxes won't look great.

openapi/components/schemas/Storefront/TransactionProcessing/ProcessingRedirect.yaml
justnero: As fas as I can tell from what we have right now this should work.

openapi/components/schemas/Storefront/TransactionProcessing/FieldTypes/ProcessingFormNumberField.yaml
justnero: It is not an HTML tag all the time, it can be presented as a set of HTML tags. On the other hand it mimics Json-Schema, we kinda do this for variable quantity plans in checkout forms as well. Could you show an example of how you see it?

openapi/components/schemas/Storefront/TransactionProcessing/ProcessingAsync.yaml
justnero: For me it's both. We need to wait for some async action so our frontend should show an instruction of how to do it. At the same time we must do a specific action – query within desired interval. I understand the point of `async` not being noun, but can't really find a suitable replacement for it.

openapi/components/schemas/Storefront/TransactionProcessing/FieldTypes/ProcessingFormNumberField.yaml
slavcodev: Where is not the tag? As far I can see every "field" represent an tag, either input, select. So I would prefer to use somethign more standard. So the fields can be:
```
tag: string
attributes: object
# some addional options if any,
# however they can be also in `attributes`,
# since custom attributes are supported already a long time
options:object
```

justnero: PAN as well as the date picker could be not just one html tag. We can use combined month+year fields for expiration. On the other hand checkboxes are not represented as one tag. And overall the label itself is a separate tag.

slavcodev: - pan might be a tag `input` with attribute `framepay`
- date picker - there are different types of input for this: `week`, `year`, `month`, etc. Also the attribute `pattern` may help to specify the format of date, and frontend will decide how to render this field
- checkboxes as per HTML are different fields, so can be all in the resposen - frontend will decide how to render multiple checkboxes

PR #265, opened 2020-11-05
Add amounts property to CashToCode settings object

ninjasimon:
This allows transaction amounts to be defined as a CashToCode Gateway Account setting. Amounts may be provided as a comma-separated list of formatted numeric values, ex `'1,2.50,5,10.00'`

### DIFF
--- a/openapi/components/schemas/GatewayAccountConfig/CashToCode.yaml
+++ b/openapi/components/schemas/GatewayAccountConfig/CashToCode.yaml
@@ -35,3 +35,6 @@ allOf:
             type: boolean
             description: Skip amount selection screen
             default: false
+          amounts:
+            type: string
+            description: A comma-separated list of formatted numeric amounts to allow


openapi/components/schemas/GatewayAccountConfig/CashToCode.yaml
```diff
@@ -35,3 +35,6 @@ allOf:
             type: boolean
             description: Skip amount selection screen
             default: false
+          amounts:
+            type: string
+            description: A comma-separated list of formatted numeric amounts to allow
```
VladimirBerdnik, line 40: having `example` property would be nice I think

slavcodev:
I'm curios why comma-seprated string andn not array?

ninjasimon:
@slavcodev, @colinbird suggested adding this as a string when we discussed implementation, but I can easily change it if you would prefer. @colinbird any thoughts on the type?

colinbird:
Do we have a UI in recomm that would support an array? @slavcodev

slavcodev:
@colinbird I suppose yes https://github.com/Rebilly/api-definitions/blob/cbff77285a43056b9b539fde2a771000c94d97a9/openapi/components/schemas/GatewayAccountConfig/Directa24.yaml#L41

PR #271, opened 2020-11-11
Rename redemption code to coupon ID

slavcodev:


openapi/components/schemas/Coupon/Coupon.yaml
```diff
@@ -1,29 +1,29 @@
 type: object
-description: Coupons and Discounts
+description: Coupons and Discounts.
 required:
   - discount
   - issuedTime
 properties:
-  redemptionCode:
```
ashkarpetin, line 7: Shouldn't it be deprecated?

slavcodev: We do not keep deprecated properties in the API definition.

ashkarpetin: Do we?

https://github.com/Rebilly/api-definitions/blob/c8db940959f0153a7fd0e1b60ad57f19a1b38cb4/openapi/components/schemas/Common/CommonKycDocument.yaml#L16

https://github.com/Rebilly/api-definitions/blob/c8db940959f0153a7fd0e1b60ad57f19a1b38cb4/openapi/components/schemas/Common/CommonTransactionRequest.yaml#L36

slavcodev: This is by purpose, mistakenly or just very old definition. This is not very useful, we rather want to force people to migrate. The `deprecate` flag in OAS is useful only for code generation or when definiions are used in applications. 
Just for docs (our case) is useless and even make more harm than good.

slavcodev: I think this was discussed here in PR, should I find the PR to confirm?

ashkarpetin (approved):

colinbird (approved):

PR #272, opened 2020-11-12
Add an Experian Check Performed event

VladimirBerdnik:


openapi/components/requestBodies/Webhooks/ExperianCheckPerformed.yaml
```diff
@@ -0,0 +1,39 @@
+content:
+  application/json:
+    schema:
+      properties:
+        customerId:
+          description: The ID of the checked customer.
+          type: string
+        outcome:
+          description: The check decision where `1` - record matches, `0` - no matches found, `-1` - record mismatches.
```
XaosSintez, line 9: Does it make sense to make enum here? I also wonder if it makes sense to have string enum here, like `record matches`, `no-matches`. If not - integer enum would be good too?

VladimirBerdnik: I decided to keep original Experian value here and not to map it to string representations. Enum should be fine, will do, thanks

XaosSintez (approved):

ashkarpetin (approved):

openapi/webhooks/experian-check-performed.yaml
```diff
@@ -0,0 +1,9 @@
+summary: Experian check performed
+operationId: experian-check-performed
+tags:
+  - Customers
```
adamaltman, line 4: Seems like it should be more in the KYC tag?

VladimirBerdnik: Technically the event is fired on the customer as we have no other entity to link the event with

PR #288, opened 2020-11-26
Add Transaction Query API

slavcodev:


openapi/paths/transactions@{id}@query.yaml
```diff
@@ -0,0 +1,48 @@
...
+        Rate-Limit-Remaining:
+          $ref: ../components/headers/Rate-Limit-Remaining.yaml
+        Rate-Limit-Reset:
+          $ref: ../components/headers/Rate-Limit-Reset.yaml
+      content:
+        application/json:
+          schema:
+            $ref: ../components/schemas/Transactions/Transaction.yaml
+    '201':
+      description: Transaction query was done successfully and parsed.
```
VladimirBerdnik, line 31:
```suggestion
      description: Transaction query was done successfully and the transaction was reconciled.
```
Do you think such description may be appropriate?

openapi/components/schemas/Transactions/TransactionQuery.yaml
```diff
@@ -0,0 +1,54 @@
+type: object
+properties:
+  processResponse:
+    description: >
+      Whether the query response from payment gateway should be processed,
+      or only sent request to gateway to read the latest state.
+      If 'true' the response will be parsed and may update the transaction.
```
VladimirBerdnik, line 7:
```suggestion
      If 'true' the response will be parsed and may update or reconcile the transaction.
```

openapi/components/schemas/Transactions/TransactionQuery.yaml
```diff
@@ -0,0 +1,56 @@
...
+      The completed transaction (approved, declined, etc) will be reconciled,
+      if discrepancies will be found.
```
adamaltman, line 9: The completed transaction should be reconciled regardless -- being reconciled is different from being marked as having a discrepancy.

openapi/components/schemas/Transactions/TransactionQuery.yaml
```diff
@@ -0,0 +1,56 @@
+type: object
+properties:
+  processResponse:
```
adamaltman, line 3: Is there really any case where we do not want to process the response? 

I think we would not process the response in particular cases anyway:

- can't be understood/parsed properly

openapi/components/schemas/Transactions/TransactionQuery.yaml
```diff
@@ -0,0 +1,56 @@
...
+      or only sent request to gateway to read the latest state.
+      If 'true' the response will be parsed and may update the transaction.
+      The completed transaction (approved, declined, etc) will be reconciled,
+      if discrepancies will be found.
+    type: boolean
+    default: false
+  logRecordId:
+    description: The communication log record ID.
+    allOf:
+      - $ref: ../ResourceId.yaml
```
adamaltman, line 15: Can you refresh my memory? Is this `readOnly`?

openapi/components/schemas/Transactions/TransactionQuery.yaml
```diff
@@ -0,0 +1,56 @@
...
+  gatewayResponse:
```
adamaltman, line 16: Is this after the query?

adamaltman (changes requested):
I think we should not have a query API. Instead, I think we should have a reconcile API.

openapi/components/schemas/Transactions/TransactionQuery.yaml
slavcodev: This requires change or backend or change our language I think. Right now we create an record of "reconciliation" when the reconciliation is created by merchant or by Rebilly if we change status of completed transaction. AFAIK.

If we plan to create new instance of "reconciliation" for every "query" it requires another project, I think.
In this RFC to implement Query API (mainly for for mama API, but can be added on Recomm too) with no great change in backend behavior.

slavcodev: What do you suggest? Should we postpone this RFC and investigate what to do with reconciliation on query or maybe anything else?

openapi/components/schemas/Transactions/TransactionQuery.yaml
slavcodev: I guess, my text and naming is not clear but I don't know how to improve it. Let me explain, by "processResponse" I mean finalize transaction wiht this query. So the answer for 
> Is there really any case where we do not want to process the response?

is the example when we want to query for preview, like we do now in mama: we query, but finalize only on send request "Update transaction".

openapi/components/schemas/Transactions/TransactionQuery.yaml
slavcodev: No.

openapi/components/schemas/Transactions/TransactionQuery.yaml
slavcodev: Yes, this is the query response parsed and shows how it affects the transaction if `processResponse` is true.

openapi/components/schemas/Transactions/TransactionQuery.yaml
adamaltman: I find this confusing then. Who will supply this id? And why?

openapi/components/schemas/Transactions/TransactionQuery.yaml
adamaltman: It seems it would be better to have a separate operation for Update transaction now. Why support a flag property if we don't need it?

openapi/components/schemas/Transactions/TransactionQuery.yaml
adamaltman: I suggest:

1. Renaming: query is unclear and `query-gateway` may be more clear -- we do not want someone to implement this API to query the transaction status.

2. Removing the flag property which updates the transaction.

3. Making the log id `readOnly`.

4. Adding a new update transaction operation which may accept a log id as an input property. As mentioned, there are other reasons people may need to update the transaction result. I think we should allow them to do it as long as it creates an immutable record (marks it as having had a discrepancy and log to the timeline who, what and when).

slavcodev:
Closing this, it's incorrect.

PR #290, opened 2020-11-30
Add isTrialOnly for subscriptions

lexerom:
Add isTrialOnly param for subscriptions

### DIFF
--- a/openapi/components/schemas/Subscription/OrderTypes/subscription-order.yaml
+++ b/openapi/components/schemas/Subscription/OrderTypes/subscription-order.yaml
@@ -38,7 +38,7 @@ allOf:
             type: string
             format: date-time
       isTrialOnly:
-        description: isTrialOnly determines if subscription will be ended after trial
+        description: Whether a subscription churns after a trial period.
         type: boolean
         default: false
       invoiceTimeShift:
--- a/openapi/components/schemas/Subscription/OrderTypes/subscription-order.yaml
+++ b/openapi/components/schemas/Subscription/OrderTypes/subscription-order.yaml
@@ -38,7 +38,7 @@ allOf:
             type: string
             format: date-time
       isTrialOnly:
-        description: Whether a subscription churns after a trial period.
+        description: Whether a subscription ends after a trial period. Recurring settings are ignored if it's `true`.
         type: boolean
         default: false
       invoiceTimeShift:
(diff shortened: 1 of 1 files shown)


removed-account-123 (dismissed):

openapi/components/schemas/Subscription/OrderTypes/subscription-order.yaml
```diff
@@ -37,6 +37,10 @@ allOf:
             description: The time the trial should end
             type: string
             format: date-time
+      isTrialOnly:
+        description: isTrialOnly determines if subscription will be ended after trial
```
slavcodev, line 41:
```suggestion
        description: Whether a subscription should be ended after trial.
```

lexerom: Agree.

VladimirBerdnik (dismissed):

slavcodev (dismissed):

creocoder (dismissed):

openapi/components/schemas/Subscription/OrderTypes/subscription-order.yaml
```diff
@@ -37,6 +37,10 @@ allOf:
             description: The time the trial should end
             type: string
             format: date-time
+      isTrialOnly:
+        description: Whether a subscription churns after a trial period.
```
2e3s, line 41:
```suggestion
        description: Whether a subscription ends after a trial period.
```

2e3s:
```suggestion
        description: Whether a subscription ends after a trial period. Recurring settings are ignored if it's `true`.
```
No precision is excessive :slightly_smiling_face:

lexerom:
I am thinking about if we add new enum param like 'recurringType' and values 'trial-only', 'recurring-only' and 'default' (or absent or null) and get rid of trial and isTrialOnly params and add trialEndTime instead. So, it will be like this:
```
recurringType: 'trial-only',
trialEndTime: 'some-valid-date'
```
instead of
```
trial: {
  enabled: true,
  endTime: 'valid-date'
},
isTrialOnly: true
```
AND

`recurringType: 'recurring-only',`

instead of
```
trial: {
  enabled: false,
  endTime: null
}
```

creocoder:
@lexerom Looks overcomplicated. I think it better to add `isRecurringOnly` and use `trial.enabled` as BC property to fix consistency.

creocoder (approved):

VladimirBerdnik (approved):

PR #292, opened 2020-12-02
Add banktransfer for IPayOptions

colinbird:
This is the default value but it is confusing to have to configure "empty" instead

### DIFF
--- a/openapi/components/schemas/GatewayAccountConfig/IpayOptions.yaml
+++ b/openapi/components/schemas/GatewayAccountConfig/IpayOptions.yaml
@@ -42,7 +42,8 @@ allOf:
               - SecureHosted
           cardType:
             type: string
-            description: Manually set the card_type for iDEAL.
+            description: Manually set the card_type
             enum:
               - ideal
               - idealqr
+              - banktransfer


openapi/components/schemas/GatewayAccountConfig/IpayOptions.yaml
```diff
@@ -42,7 +42,8 @@ allOf:
               - SecureHosted
           cardType:
             type: string
-            description: Manually set the card_type for iDEAL.
+            description: Manually set the card_type
             enum:
               - ideal
               - idealqr
+              - banktransfer
```
dara123, line 49: do we have eps too?

colinbird: EPS uses "banktransfer" 

> The "card_type" parameter for both the iDeal and EPS transactions must be set to "banktransfer" instead of "ideal" currently.

colinbird: It seems we don't use `ideal` or `idealqr` anymore.  But I wanted to add this one because it's confusing to as the merchant to configure it as <blank> so as to use the default.

colinbird: Maybe we should just remove this setting.

dara123 (approved):

PR #299, opened 2020-12-10
Allow to update DCC outcome in transaction

slavcodev:


openapi/components/schemas/Transactions/Transaction.yaml
```diff
@@ -134,16 +135,18 @@ allOf:
...
-            description: The amount of markup translated to USD
             type: number
+            description: The amount of markup translated to USD
+            readOnly: true
             format: double
           outcome:
             type: string
             description: Dynamic Currency Conversion outcome
             enum:
+              - unknown
```
adamaltman, line 146: why not in alpha order?

slavcodev: I don't know. I put it in the lifecycle order from initial to final. I can change it, if you think the alpha order is better.

PR #300, opened 2020-12-10
Add Invoice.paymentFormUrl so that anyone can pay the invoice

2e3s:


openapi/components/schemas/Common/CommonInvoice.yaml
```diff
@@ -118,6 +118,13 @@ properties:
     type: integer
     description: Collection period - difference between paidTime and issuedTime in days.
     readOnly: true
+  paymentFormUrl:
+    type: string
+    readOnly: true
```
slavcodev, line 123: It may be wrietable, just in case the merchant hosted this form.

2e3s: Yeah, but it would need more efforts from us to store this, as opposed to e.g. be dynamically generated. IMO if it's to be done quickly, it's better stay read only for now.

slavcodev: Hmm, the URL to Storefront UI seems require the JWT, so I suppose the URL will be persisted anyway.

2e3s:
```suggestion
```

2e3s: Writeable field would be implemented later.

slavcodev (approved):

iamdevice (approved):

PR #313, opened 2020-12-24
Replace Subscription to Order where it's not about recurring orders

2e3s:
This changes only descriptions and more importantly tags, so it's only about what people see in UI (thus unrelated to backward compatibility). The reason is that when you're looking to implement something like a cart and browsing the API, you would overlook the "Subscriptions" subdivision.

This would introduce a certain inconsistancy with API resources and endpoints though as API remains to be `/subscriptions/`.

https://preview.redoc.ly/rebilly-core/subscription-to-order/tag/Orders

### DIFF
--- a/openapi/components/schemas/Common/CommonTransaction.yaml
+++ b/openapi/components/schemas/Common/CommonTransaction.yaml
@@ -101,13 +101,13 @@ properties:
     items:
       $ref: ../ResourceId.yaml
   subscriptionIds:
-    description: The subscription IDs related to transaction's invoice(s).
+    description: The orders IDs related to transaction's invoice(s)
     readOnly: true
     type: array
     items:
       $ref: ../ResourceId.yaml
   planIds:
-    description: The plan IDs related to transaction's subscription(s).
+    description: The plan IDs related to transaction's order(s).
     readOnly: true
     type: array
     items:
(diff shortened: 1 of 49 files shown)


openapi/components/schemas/Common/CommonTransaction.yaml
```diff
@@ -101,13 +101,13 @@ properties:
     items:
       $ref: ../ResourceId.yaml
   subscriptionIds:
-    description: The subscription IDs related to transaction's invoice(s).
+    description: The orders IDs related to transaction's invoice(s)
```
VladimirBerdnik, line 104:
```suggestion
    description: The orders IDs related to transaction's invoice(s).
```

openapi/paths/subscriptions@{id}@change-plan.yaml
```diff
@@ -3,16 +3,15 @@ parameters:
...
+    - Orders
+  summary: Change an order's plan
   operationId: PostSubscriptionPlanChange
   description: >
-    Change a subscription's plan and designate when and if there should be pro.
-    rata credits given.
+    Change an order's plan and designate when and if there should be pro-rata credits given.
 
-    Only active subscriptions with a single plan can be changed.
+    Only active order orders with a single plan can be changed.
```
VladimirBerdnik, line 12:
```suggestion
    Only active subscription orders with a single plan can be changed.
```

slavcodev (approved):

VladimirBerdnik (approved):

PR #316, opened 2021-01-07
fix: this should be anyOf not oneOf

adamaltman:


slavcodev:
I think this is incorrect. We do not accept partial "PaymentCardCreatePlain" and "BankAccountCreatePlain"

adamaltman:
I think you're wrong. We don't need to. The issue is when validation cannot clearly distinguish between objects.

adamaltman:
I think you misunderstand what anyOf means vs. oneOf. anyOf doesn't mean "one or more" sent in a request.

slavcodev:
>  **9.2.1.2. anyOf**
>
> An instance validates successfully against this keyword if it validates successfully against at least one schema defined by this keyword's value. Note that when annotations are being collected, all subschemas MUST be examined so that annotations are collected from each subschema that validates successfully. 

> **9.2.1.3. oneOf**
>
> An instance validates successfully against this keyword if it validates successfully against exactly one schema defined by this keyword's value. 

https://json-schema.org/draft/2019-09/json-schema-core.html

IAs far I understand the `anyOf`, `oneOf` and `allOf` is not logical operators like `OR`/`AND`,
they are also the validation keywords with different validation meaning. 
They are used for composition.
For instance, `anyOf` is kind of to merge of keywords from different definitions, It's kind of multple inheritance.

According specification and your changes 

```yaml
anyOf:
        - $ref: ../schemas/PaymentInstruments/PaymentInstrumentCreateToken.yaml
        - $ref: ../schemas/PaymentCards/PaymentCardCreatePlain.yaml
```

an instance in a request

````yaml
token: bank-account-token-1
customerId: customer-1
method: payment-card
pan: 4111-1111-1111-1111
expYear: 2022
expMonth: 12
billingAddress: {}
```

is valid and backend should handle this. I don't think it's correct.

I know, I'm often wrong, for example I thought we need at least 3 approvals reviews to change the public API.
But in this case I believe I'm not worng and the change in this PR is incorrect, unless you're going to change backend and handle more than one payment instrument in one request.

adamaltman:
There is nothing to change on the backend. This is documented incorrectly. 

🐼

adamaltman:
Also, according to the specification, you are incorrect, this is not the case:

> anyOf:
        - $ref: ../schemas/PaymentInstruments/PaymentInstrumentCreateToken.yaml
        - $ref: ../schemas/PaymentCards/PaymentCardCreatePlain.yaml
an instance in a request
```
token: bank-account-token-1
customerId: customer-1
method: payment-card
pan: 4111-1111-1111-1111
expYear: 2022
expMonth: 12
billingAddress: {}
```
> is valid and backend should handle this. 

This is incorrect. It is invalid. `anyOf` is about ambiguity, not about blending schemas.

slavcodev:
I've not said anything about "blending schemas" though. And probably we're reading the different specification or reading the same but i ndifferent way, because I don't see anything in it about "ambiguity". I see this in addional to `9.2.1.2` and `9.2.1.2` I mentioned before:

>  9.2.1. Keywords for Applying Subschemas With Boolean Logic
>
> These keywords correspond to logical operators for combining or modifying the boolean assertion results of the subschemas. They have no direct impact on annotation collection, although they enable the same annotation keyword to be applied to an instance location with different values. Annotation keywords define their own rules for combining such values.

So I understand that validator that follows the specifation will validate this instance:

```yaml
token: bank-account-token-1
customerId: customer-1
method: payment-card
pan: 4111-1111-1111-1111
expYear: 2022
expMonth: 12
billingAddress: {}
```

in this way

1. It will anotate as `true` by `PaymentInstrumentCreateToken`
2. it will anotate as `true` by `PaymentCardCreatePlain`
3. finally using keywords `anyOf` will annotate the result as `true` because more than one previous flags are `true`

Notice, it's not about "blending schemas", it is about validated instance against each schema. Specificaiton clearly claims in `9.2.1.2. anyOf` that the validator MUST examinate all the schemas and annotate which are valid.

I would be happy to just trust your experience and knowledges even w/o any arguments. But I can't. I feel like I understand this part of specification well, and don't see which part I can misread and misunderstand.

I would appreciate a lot if you will point me or may be you have any links which canc clarify the specificaiton, and will help to understand where I'm wrong. 

🐼

PR #342, opened 2021-02-07
Query transaction API

adamaltman:
Read the full RFC for more context.

### DIFF
--- a/openapi/components/schemas/Transactions/TransactionQuery.yaml
+++ b/openapi/components/schemas/Transactions/TransactionQuery.yaml
@@ -1,0 +1,21 @@
+type: object
+properties:
+  result:
+    description: Transaction result.
+    type: string
+    readOnly: true
+    enum:
+      - abandoned
+      - approved
+      - canceled
+      - declined
+      - unknown
+  amount:
+    description: The transaction's amount.
+    type: number
+    format: double
+    readOnly: true
+  currency:
+    readOnly: true
+    allOf:
+      - $ref: ../CurrencyCode.yaml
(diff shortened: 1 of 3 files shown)


openapi/components/schemas/Transactions/TransactionQuery.yaml
```diff
@@ -0,0 +1,21 @@
...
+      - unknown
+  amount:
+    description: The transaction's amount.
+    type: number
+    format: double
+    readOnly: true
+  currency:
+    readOnly: true
+    allOf:
+      - $ref: ../CurrencyCode.yaml
```
svensono, line 21: Would it be valuable to add the gateway responseCode and responseMessage?

openapi/components/schemas/Transactions/TransactionQuery.yaml
```diff
@@ -0,0 +1,21 @@
...
+  result:
+    description: Transaction result.
+    type: string
+    readOnly: true
+    enum:
+      - abandoned
+      - approved
+      - canceled
+      - declined
+      - unknown
```
colinbird, line 12: I am thinking of the `waiting-capture` scenario.   Perhaps it should be added here, or both `result`/`status` returned.

viceromag (approved):

XaosSintez (approved):

openapi/components/schemas/Transactions/TransactionQuery.yaml
adamaltman: Yes

PR #344, opened 2021-02-09
Fix Paymero settings

dara123:
Seem like Recomm can't handle Ref in Gateway account settings


mattsanf (approved):

ashkarpetin (approved):

slavcodev:
@dara123 have you reported the but in Recomm? There are more gateway config definitions with this kind of schema.

dara123:
Digi need to test this soon so It's easier to fix it like this. will create issue for Recomm to fix it later

PR #355, opened 2021-02-22
Add overwrite to KYC document documentMatches

subenson:


openapi/components/schemas/KycDocument/IdentityMatches.yaml
```diff
@@ -0,0 +1,42 @@
+description: Identity match properties
```
svensono, line 1:
```suggestion
description: Identity match properties.
```

openapi/components/schemas/KycDocument/ProofOfIdentity.yaml
```diff
@@ -1,58 +1,29 @@
...
+            example: Doe
+          dateOfBirth:
+            description: The date of birth found on the document, null if not found.
+            allOf:
+              - $ref: ../ServerTimestamp.yaml
+      documentMatches:
+        readOnly: true
+        allOf:
+          - $ref: ./IdentityMatches.yaml
+      overwriteData:
```
adamaltman, line 24: This part doesn't make sense to me.

openapi/components/schemas/KycDocument/ProofOfIdentity.yaml
```diff
@@ -1,58 +1,29 @@
 allOf:
   - $ref: ../Common/CommonKycDocument.yaml
   - properties:
-      documentMatches:
+      customerInfo:
```
adamaltman, line 4: I don't like the customerInfo in this API. What happens when the customer info changes? Does the GetKycDocument API change automatically too? Does the `updatedTime` change? Etc... I don't think it makes sense. The user can use the customers API to get the customers data.

openapi/components/schemas/KycDocument/ProofOfIdentity.yaml
adamaltman: It makes sense to me now.

adamaltman, line 24:
```suggestion
      documentMatches:
        allOf:
          - $ref: ./IdentityMatches.yaml
      parsedData:
        readOnly: true
```

I propose a minor change. They can edit the `documentMatches`. They have code that reads that and that's what they want to use. Instead, they can find the original data in the `parsedData` object.

Another possibility, is this should actually be a separate API:
```
POST /kyc-documents/{id}/documentMatches
{
   "firstName": "Adam",
   "lastName": "Altman", etc...

}

```

openapi/components/schemas/KycDocument/ProofOfIdentity.yaml
svensono: You're right! I'll remove it

openapi/components/schemas/KycDocument/ProofOfIdentity.yaml
svensono: I like your proposition of the parsedData. As for the seperate API, would this eliminate the documentMatches in the GetKycDocument endpoint, or does it remain as it is, and this endpoint is added for updating?

adamaltman: It would remain as-is. This other API is exclusively for updating document matches data.

adamaltman (dismissed):

openapi/components/schemas/KycDocument/IdentityMatches.yaml
```diff
@@ -0,0 +1,44 @@
...
+    description: The customer first name if it was matched, null otherwise.
+    type: string
+    example: John
+  lastName:
+    description: The customer last name if it was matched, null otherwise.
+    type: string
+    example: Doe
+  dateOfBirth:
+    description: The date of birth found on the document, null if not found.
+    readOnly: false
```
ashkarpetin, line 25: Isn't it readonly = false by default?

justnero: It is, but in `ServerTimestamp.yaml` it's `readOnly: true` so to make timestamp writable I've overwritten it here.

ashkarpetin: Then what is the point to ServerTimestamp.yaml here?

ashkarpetin (approved):

openapi/components/schemas/KycDocument/IdentityMatches.yaml
adamaltman: Yes, this isn't really overwriting it either. Please describe it. Also, is date of birth date-time or just date?

openapi/components/schemas/KycDocument/IdentityMatches.yaml
```diff
@@ -0,0 +1,41 @@
...
+    type: string
+    example: John
+  lastName:
+    description: The customer last name if it was matched, null otherwise.
+    type: string
+    example: Doe
+  dateOfBirth:
+    description: The date of birth found on the document, null if not found.
+    type: string
+    format: date-time
```
adamaltman, line 26: Is this a date-time or just a date?

justnero: It's date-time, `Y-m-d\TH:i:sP` as far as I can tell.

iamdevice (approved):

PR #356, opened 2021-02-23
Propose event rule sets

justnero:
Still missing operation descriptions, rules/binds reordering and rule-set changes API.

### DIFF
--- a/openapi/components/schemas/Rules/Bind.yaml
+++ b/openapi/components/schemas/Rules/Bind.yaml
@@ -1,6 +1,10 @@
 type: object
 description: The rule.
 properties:
+  id:
+    readOnly: true
+    allOf:
+      - $ref: ../ResourceId.yaml
   name:
     type: string
   labels:
--- a/openapi/components/schemas/Rules/DraftRuleSet.yaml
+++ b/openapi/components/schemas/Rules/DraftRuleSet.yaml
@@ -1,20 +1,9 @@
 type: object
-description: RuleSet for particular event.
+description: Draft rule set for particular event.
 properties:
   id:
     description: RuleSet identifier string.
     readOnly: true
-    allOf:
-      - $ref: ../ResourceId.yaml
-  isActive:
-    type: boolean
-    readOnly: true
-  version:
-    type: string
-    readOnly: true
-  parent:
-    description: The parent RuleSet identifier string.
-    nullable: true
     allOf:
       - $ref: ../ResourceId.yaml
   name:
@@ -23,14 +12,12 @@ properties:
   description:
     type: string
     description: The description of the rule set contents.
-  binds:
-    type: array
-    items:
-      $ref: ./Bind.yaml
-  rules:
-    type: array
-    items:
-      $ref: ./Rule.yaml
+  versionHash:
+    type: string
+    description: The rule set version hash.
+    nullable: true
+  createdTime:
+    $ref: ../ServerTimestamp.yaml
   updatedTime:
     $ref: ../ServerTimestamp.yaml
   _links:
--- a/openapi/components/schemas/Rules/EventRuleSet.yaml
+++ b/openapi/components/schemas/Rules/EventRuleSet.yaml
@@ -1,0 +1,42 @@
+type: object
+description: RuleSet for particular event.
+properties:
+  id:
+    description: RuleSet identifier string.
+    readOnly: true
+    allOf:
+      - $ref: ../ResourceId.yaml
+  isActive:
+    type: boolean
+    readOnly: true
+  version:
+    type: string
+    readOnly: true
+  parent:
+    description: The parent RuleSet identifier string.
+    nullable: true
+    allOf:
+      - $ref: ../ResourceId.yaml
+  name:
+    type: string
+    description: The name of the rule set.
+  description:
+    type: string
+    description: The description of the rule set contents.
+  binds:
+    type: array
+    items:
+      $ref: ./Bind.yaml
+  rules:
+    type: array
+    items:
+      $ref: ./Rule.yaml
+  updatedTime:
+    $ref: ../ServerTimestamp.yaml
+  _links:
+    type: array
+    description: The links related to resource.
+    readOnly: true
+    minItems: 1
+    items:
+      $ref: ../Links/SelfLink.yaml
--- a/openapi/components/schemas/Rules/RuleSetConfig.yaml
+++ b/openapi/components/schemas/Rules/RuleSetConfig.yaml
@@ -3,16 +3,14 @@ description: Set of rules for particular event.
 properties:
   versionHash:
     type: string
-    readOnly: true
-  parentHash:
-    type: string
-    readOnly: true
   binds:
     type: array
+    readOnly: true
     items:
       $ref: ./Bind.yaml
   rules:
     type: array
+    readOnly: true
     items:
       $ref: ./Rule.yaml
   updatedTime:
@@ -25,4 +23,4 @@ properties:
     items:
       $ref: ../Links/SelfLink.yaml
 required:
-  - rules
+  - versionHash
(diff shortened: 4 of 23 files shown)


openapi/paths/events@{eventType}@rule-sets@{id}@apply.yaml
```diff
@@ -0,0 +1,25 @@
+parameters:
+  - $ref: ../components/parameters/organizationId.yaml
+  - $ref: ../components/parameters/systemEventType.yaml
+  - $ref: ../components/parameters/resourceId.yaml
+post:
+  tags:
+    - Rules
+  summary: Apply a rule set for event.
```
slavcodev, line 8: What does this operation mean?

openapi/components/schemas/Rules/EventRuleSet.yaml
```diff
@@ -0,0 +1,42 @@
+type: object
+description: RuleSet for particular event.
+properties:
+  id:
+    description: RuleSet identifier string.
+    readOnly: true
+    allOf:
+      - $ref: ../ResourceId.yaml
+  isActive:
```
slavcodev, line 9: Is this flag supposed for drafts and publiched set?

justnero: Yes, endpoint you've asked above is right about making one rule-set active.

slavcodev: I think we can just patch this proeprty, I don't understand the necessity of the extra endpoint.

slavcodev: Also I have an concern about the version, how it correlates with "drafts" and new API. If I make draft from set with version 5 and publish it after 1-2 months, which version it will be? Let say the original was changed too.

slavcodev: Another question about the order and the behavior of executing of the multiple rules sets?
For example, I have 2 sets of "Transaction process was requested" with different actions "pick gateway account". Which action takes precedent?

justnero: >I think we can just patch this proeprty, I don't understand the necessity of the extra endpoint.

Frontend implementations tend to send the whole object to the backend on patch even if only one field has changed. Imagine multiple users at the same time would save their rule-set with `isActive: true` just because one selected it to be `true`, but the other just had it as `true`. Extra endpoint just ensures this action was 100% intentional and is not a mistake in using the API mainly because changing live rules may be quite dangerous.

justnero: >Also I have an concern about the version, how it correlates with "drafts" and new API. If I make draft from set with version 5 and publish it after 1-2 months, which version it will be? Let say the original was changed too.

I would probably make active rule-sets immutable. Version is a hash of the latest change made to this rule-set, I was thinking about parenting based on rules hash like a commit hash while rule-set is a branch HEAD. This part is still draft.

justnero: >Another question about the order and the behavior of executing of the multiple rules sets?
For example, I have 2 sets of "Transaction process was requested" with different actions "pick gateway account". Which action takes precedent?

Only one rule-set for event can be active at a time. Rule set order doesn't really matter, but the active one is always displayed first just for convenience and better UX. I have a work to do about designing actual rules/binds reordering in a set, no idea how to do it in a RESTful way, would appreciate any help on this one.

slavcodev: > Extra endpoint just ensures this action was 100% intentional 

It's not RESTful, but okay.

slavcodev: > I would probably make active rule-sets immutable. 
Active set is immutable, no problem with this, this is why we have versions. I think my question was misunderestood.
Imagine someone branches a draft from a set with version 5. The draft is edited a few times (it's muttable, we don';t increase versions). In the same time, the original set was changed a few times and become version 10.

What version will be draft after publishing? I imagine we will assign the next version based in `parent`? But the parent will be version 5 or version 10?

slavcodev: > Only one rule-set for event can be active at a time.

Hmm, I don't understand why we create collection of set then. :(

justnero: As you saw versions are strings this time, so I'm really on my way to rule-set git. This numeric versioning makes more harm than use – our clients have versions 10k+ and that doesn't make any sense. What make sense is the ability to see what have changed and where it came from.

creocoder: @justnero So the idea is to absorb numeric versions with hash? I like that.

openapi/components/schemas/Rules/DraftRuleSet.yaml
```diff
@@ -0,0 +1,29 @@
...
+    readOnly: true
+    allOf:
+      - $ref: ../ResourceId.yaml
+  name:
+    type: string
+    description: The name of the rule set.
+  description:
+    type: string
+    description: The description of the rule set contents.
+  versionHash:
```
slavcodev, line 15: just `hash` would not be enough?

openapi/components/schemas/Rules/DraftRuleSet.yaml
```diff
@@ -0,0 +1,29 @@
+type: object
+description: Draft rule set for particular event.
+properties:
+  id:
+    description: RuleSet identifier string.
```
slavcodev, line 5:
```suggestion
    description: Draft identifier string.
```

openapi/components/schemas/Rules/Bind.yaml
```diff
@@ -1,6 +1,10 @@
 type: object
 description: The rule.
 properties:
+  id:
```
slavcodev, line 4: Can you elabborate why we need ID for rule and bind?

openapi/components/schemas/Rules/RuleSetConfig.yaml
```diff
@@ -0,0 +1,26 @@
+type: object
+description: Set of rules for particular event.
+properties:
+  versionHash:
```
slavcodev, line 4: Wait? This is writable?

openapi/components/schemas/Rules/RuleSetVersion.yaml
```diff
@@ -12,7 +15,7 @@ properties:
     type: array
     items:
       $ref: ./Rule.yaml
-  createdTime:
+  updatedTime:
```
slavcodev, line 18: Version is immutable, isn't it? How it can be "updated"?

openapi/components/schemas/Rules/RuleSetConfig.yaml
```diff
@@ -0,0 +1,26 @@
+type: object
```
slavcodev, line 1: How this differes from existing `RuleSet`, why we need new resource and new paths?

openapi/components/schemas/Rules/DraftRuleSet.yaml
creocoder: I would say `version` may be enough, but only `hash`? Hash is just minor technical detail while version represent nature of property. Version may be represented by number or by hash or sometimes by time. I don't think it good idea to name properties by its type.

slavcodev: There is difference between `version` and `hash` meaning though. In this case it's more `fingerprint`/`hash` than `version`. The versions are usually something which can be orderer in row, you're right it can be numbers, time, and sometimes set of chars, but those chars usually addional info, isn't it?

openapi/components/schemas/Rules/Bind.yaml
2e3s: To track **trivially** rearrangements and changes in general.

PR #361, opened 2021-03-01
Add Organization questionnaire

justnero:


openapi/components/schemas/OrganizationQuestionnaire.yaml
```diff
@@ -0,0 +1,25 @@
+type: object
+required:
```
adamaltman, line 2: The fields should be optional. Nothing should be required here.

openapi/components/schemas/OrganizationQuestionnaire.yaml
```diff
@@ -0,0 +1,25 @@
...
+  - role
+  - monthlyTransactions
+  - products
+  - integrationType
+  - timeToGoLive
```
adamaltman, line 7: Rename to launchTiming

adamaltman (changes requested):

adamaltman (approved):

iamdevice (approved):

PR #364, opened 2021-03-08
Add Drafts Rules API

2e3s:
It is intentionally simplified to be a bare minimum over the existing APIs to satisfy the required functionality. Frontend will just use the Rules Drafts as a source for actual rules when someone wants to clone or apply a draft rule set. Given the functionality, any other API on top of this would be just a sugar-wrapper doing the same things (maybe moving responsibilities here and there).

[Preview](https://preview.redoc.ly/rebilly-users/draft-rule-set/tag/Rules)


openapi/components/schemas/Rules/Bind.yaml
```diff
@@ -1,6 +1,12 @@
 type: object
 description: The rule.
 properties:
+  id:
+    description: |
+      The rule's unique identifier which defines whether an event is emitted
+      when the rule is created, updated or removed.
```
2e3s, line 7: This is the main difference from #356 even though it existed there as well. It allows to do all the timeline work in the backend, and the frontend can do its diff work based on this as well. Hence all other APIs from #356 are simply unnecessary.

openapi/components/schemas/Rules/RuleSetDraft.yaml
```diff
@@ -0,0 +1,48 @@
...
+      - $ref: ../ResourceId.yaml
+  version:
+    type: integer
+    readOnly: true
+    description: |
+      The version of the active rule set when the draft was created.
+      This can be used to know how further the active rule set has advanced since the draft was created.
+  binds:
+    type: array
+    description: All binds are executed regardless other binds, there can't be a final bind.
```
adamaltman, line 17: Regardless "of" other binds (missing of).

adamaltman (changes requested):
Minor change... missing "of" in a few copy/pasted descriptions.

openapi/components/schemas/Rules/RuleSetDraft.yaml
```diff
@@ -0,0 +1,48 @@
+type: object
+description: Rules draft for a particular event.
+properties:
+  id:
+    type: string
+    readOnly: true
+    allOf:
+      - $ref: ../ResourceId.yaml
+  version:
```
slavcodev, line 9: I think we discussed to not save draft versions. What is purpose of the draft version?

openapi/components/schemas/Rules/RuleSetDraft.yaml
```diff
@@ -0,0 +1,48 @@
...
+  updatedTime:
+    $ref: ../ServerTimestamp.yaml
+  _links:
+    type: array
+    description: The links related to resource.
+    readOnly: true
+    items:
+      $ref: ../Links/SelfLink.yaml
+required:
+  - rules
```
slavcodev, line 48: What about name?

2e3s: If the user doesn't want to specify it then they don't specify IMO

slavcodev: How they will distinguish drafts w/o names?

openapi/components/schemas/Rules/RuleSetDraft.yaml
2e3s: To know that the live ruleset has advanced from the draft ruleset. Since it can be computed from Rules History API (by comparing the live version updatedTime and the draft's created time), it's only a materialized link to the ruleset version which was actual when the draft was created.

2e3s: It could be named something like "lastActiveVersion"

slavcodev: Can you please elaaborate, not sure I unerstood your answer. Can you explain on this scenario

1) My active rules set has version 2
2) I clone a draft "A" (is it version 2?)
3) I clone another draft "B" (is it version 2?)
4) I publish "A" as active (it will be version 3?)
5) I publish "B" as active (what version will be it now?)

slavcodev: Another story

1) My active rules set has version 2
2) I clone a draft (as far I understand draft will have version 2 too, as active set from which it was cloned)
3) I changed the draft - will it keep `version = 2`? 

How it will be useful? To compare this draft with version 2 from history? Does it make sense? Should not we always ocmpare with active rules set, whatever version it is?

adamaltman (dismissed):
I reject this change. I think the version should be clarified a bit, maybe renamed to `clonedVersion`

openapi/components/schemas/Rules/RuleSetDraft.yaml
adamaltman: I disagree with @slavcodev as I mentioned. Please re-read the RFC. I think this property could be named better as `clonedVersion` or similar.

adamaltman: > Should not we always compare with active rules set, whatever version it is?

Yes, but from which version was it cloned?

2e3s: A draft can be cloned from 
- an older rule set,
- an active rule set
- another draft 
- or a new draft can be created from the scratch

The API is agnostic to that.
I have thought about having a writeable "version" to accomodate all of that. Hence this property will be used only by the frontend with no underlying logic behind.

openapi/components/schemas/Rules/RuleSetDraft.yaml
```diff
@@ -0,0 +1,52 @@
...
+  - name
+  - version
```
slavcodev, line 52:
```suggestion
  - clonedVersion
```
?

openapi/components/schemas/Rules/RuleSetDraft.yaml
slavcodev: > I think this property could be named better as clonedVersion or similar.

I propose `baseVersion`. Since cloned seems not appropriate, the draft could be posted with different some or all rules, taken another set as base. It does not seem a real "clone".

2e3s: I agree, I wanted to suggest something different from "clone" but I didn't come up with a better term

2e3s:
devs please review

2e3s:
devs please review

ashkarpetin (approved):

2e3s:
devs please review

2e3s:
devs please review

2e3s:
devs please review

ghost (approved):
Looks good. The `RuleSetDraft` has some duplicates from `RuleSet`, e.g. `rules` and `binds` though. We should consider to refactor the definitions in order to DRY.

2e3s:
Bind and rules are on a the few same fields and they are already separate in `Bind,yaml`. I've thought that DRY would be overcomplication here.

slavarebilly:
It's overcomplication to repeat your self and keep documentation consistent for the things that represent the same. IMO.

PR #365, opened 2021-03-08
Fix event rule operation ID

ashkarpetin:


p3yman (approved):

openapi/paths/events@{eventType}@rules.yaml
```diff
@@ -34,7 +34,7 @@ put:
   tags:
     - Rules
   summary: Update the rules for event
-  operationId: PutEventRuleCollection
+  operationId: PutEventRule
```
slavcodev, line 37: I think this might be confusing, since the operation works with rules set and not one single rule.

ashkarpetin: @2e3s Please check this

https://github.com/Rebilly/api-definitions/pull/364/files#diff-c0ec58b81489a48627f8f1c6b4f570c7d70ac5265736e3dbbddf58dcd4698123R59

ashkarpetin: @slavcodev Are you suggesting to leave `PutEventRuleCollection`?

slavcodev (approved):

openapi/paths/events@{eventType}@rules.yaml
2e3s: I don't see how it's inconsistent and affects my PR

PR #332, opened 2021-01-25
Add tolerance and targetCurrency setting to Paymero

dara123:

PR #368, opened 2021-03-12
Add tolerance percentage to Piastrix

dara123:


openapi/components/schemas/GatewayAccountConfig/Piastrix.yaml
```diff
@@ -19,5 +19,15 @@ allOf:
...
           - secretKey
+      settings:
+        type: object
+        description: Piastrix settings object.
+        properties:
+          tolerancePercentage:
+            type: integer
+            description: Tolerance percentage for settled amount.
+            minimum: 0
+            maximum: 5
```
ashkarpetin, line 30: I wonder why it is different compared to Paymero https://github.com/Rebilly/api-definitions/pull/332/

dara123: we fixed Paymero. should be max 5

dara123: https://github.com/Rebilly/api-definitions/blob/master/openapi/components/schemas/GatewayAccountConfig/Paymero.yaml#L28

ashkarpetin (approved):

slavcodev (commented):
Please re-use definition https://github.com/Rebilly/api-definitions/blob/6db1f07d41faea89059edd5e1b5825856b8fdb3a/openapi/components/schemas/GatewayAccountConfig/UPayCard.yaml#L33

dara123:
I don't think that work yet. I still see bug issue for this open

adamaltman (approved):

openapi/components/schemas/GatewayAccountConfig/Piastrix.yaml
```diff
@@ -19,5 +19,15 @@ allOf:
...
+            example: 5
```
adamaltman, line 31:
```suggestion
           $ref: ../Gateways/AmountAdjustmentTolerance.yaml
```

adamaltman (changes requested):

openapi/components/schemas/GatewayAccountConfig/Piastrix.yaml
dara123: Using ref doesn't work on Gateway account setting now. I open ticket for Recomm to fix this. I will change to ref after they fix that.

ashkarpetin (approved):

adamaltman (approved):

PR #372, opened 2021-03-16
Fix collection discrepancies

agualis:
Adding missing parameters to multiple collection operations.

### DIFF
--- a/openapi/paths/search.yaml
+++ b/openapi/paths/search.yaml
@@ -9,11 +9,15 @@ get:
     Search merchant's data to return resources such as customers, invoices,
     orders, transactions.
   parameters:
+    - $ref: ../components/parameters/collectionSort.yaml
+    - $ref: ../components/parameters/collectionLimit.yaml
+    - $ref: ../components/parameters/collectionOffset.yaml
     - name: q
       in: query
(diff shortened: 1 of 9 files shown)


ashkarpetin (approved):

justnero (approved):

openapi/paths/search.yaml
```diff
@@ -9,6 +9,9 @@ get:
     Search merchant's data to return resources such as customers, invoices,
     orders, transactions.
   parameters:
+    - $ref: ../components/parameters/collectionSort.yaml
+    - $ref: ../components/parameters/collectionLimit.yaml
+    - $ref: ../components/parameters/collectionOffset.yaml
```
adamaltman, line 14: I don't think search supports these parameters.

adamaltman: @ashkarpetin @justnero are you seeing something different from me?

justnero: @adamaltman It is supported, but rather supported per-resource and not for the whole collection at the same time. It would sort, limit and offset each resource collection. I've shared a link in Slack.

iamdevice (approved):

PR #381, opened 2021-03-19
Increase Piastrix tolerance percentage

dara123:


openapi/components/schemas/GatewayAccountConfig/Piastrix.yaml
```diff
@@ -27,7 +27,7 @@ allOf:
             type: integer
             description: Tolerance percentage for settled amount.
             minimum: 0
-            maximum: 5
+            maximum: 20
```
ashkarpetin, line 30: Why only Piastrix?

colinbird: Seems like we should change it for both

PR #428, opened 2021-04-19
Proposal for the rules diff to be calculated on the backend

rensftw:


openapi/components/schemas/Timelines/list.yaml
```diff
@@ -6,4 +6,12 @@ allOf:
         type: array
         description: Table data. Each array element represents a line.
         items:
-          type: string
+          type: object
```
XaosSintez, line 9: Is it really array of objects? Do you mean that every rule change will have it's own summary and diff?

What are items (line 17) in this case?

XaosSintez:
I don't like that timeline table will have something that specific. If we change it to `ruleset-diff` type (or something like that) it will make more sense to me

rensftw:
This proposal has been ruled out.

PR #432, opened 2021-04-21
Add sdk names to operations

agualis:
- Used a script to update every operation path from the info in the `SDK generator customizations` .

- Added custom lint rule to detect operations missing `x-sdk-operation-name` property.


ghost (commented):
LGTM.  However, I would suggest to add custom [rule for Redoc](https://github.com/Rebilly/api-definitions/blob/master/plugins/custom-rules.js) and assert all operations contains this keyword.

agualis:
> LGTM. However, I would suggest to add custom [rule for Redoc](https://github.com/Rebilly/api-definitions/blob/master/plugins/custom-rules.js) and assert all operations contains this keyword.

I just added it. 

If `redoc` allowed to setup lint rules to specifying which` source api-definitions` to validate,  we could avoid the validation of `reports.yaml` (we don't auto-generate SDK code for them yet) but I don't think that's possible. Instead of that,  I added custom code to skip `experimental` operations validation (against the new custom rule). I don't love it but it works.

ghost (dismissed):

openapi/paths/transactions@{id}@cancel.yaml
```diff
@@ -0,0 +1,39 @@
+parameters:
+  - $ref: ../components/parameters/resourceId.yaml
+  - $ref: ../components/parameters/organizationId.yaml
+post:
+  tags:
+    - Transactions
+  summary: Cancel a pending or suspended transaction
+  operationId: PostTransactionCancellation
```
iamdevice, line 8: As I know we removed transaction cancellation operation

agualis: Good catch. I created this PR before that deletion and now it was added again as part of the auto-merge.  I just manually deleted it again.

iamdevice (approved):

ghost (approved):

PR #444, opened 2021-04-29
Separate List's values from the resource

2e3s:
Lists collection can be very slow as the values can be very big for each list.

### DIFF
--- a/openapi/components/schemas/List.yaml
+++ b/openapi/components/schemas/List.yaml
@@ -36,4 +36,15 @@ properties:
     readOnly: true
     minItems: 1
     items:
-      $ref: ./Links/SelfLink.yaml
+      - $ref: ./Links/SelfLink.yaml
+      - type: object
+        allOf:
+          - $ref: ./Links/Link.yaml
+        properties:
+          rel:
+            description: The link type.
+            type: string
+            enum:
+              - values
+        required:
+          - rel
--- a/openapi/components/schemas/List.yaml
+++ b/openapi/components/schemas/List.yaml
@@ -18,6 +18,7 @@ properties:
     type: string
   values:
     description: List values.
+    writeOnly: true
     type: array
     items:
       type: string
--- a/openapi/paths/lists@{id}@{version}@values.yaml
+++ b/openapi/paths/lists@{id}@{version}@values.yaml
@@ -12,7 +12,7 @@ get:
   tags:
     - Lists
   summary: Retrieve values for the given list's version
-  operationId: GetListVersion
+  operationId: GetListVersionValues
   x-sdk-operation-name: getByVersion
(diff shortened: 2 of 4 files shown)


openapi/components/schemas/List.yaml
```diff
@@ -18,6 +18,7 @@ properties:
     type: string
   values:
     description: List values.
+    writeOnly: true
```
ashkarpetin, line 21: This is a breaking change for API consumers

openapi/paths/lists@{id}@{version}@values.yaml
```diff
@@ -0,0 +1,31 @@
...
+    description: List version.
+    schema:
+      type: integer
+      minimum: 1
+get:
+  tags:
+    - Lists
+  summary: Retrieve values for the given list's version
+  operationId: GetListVersionValues
+  x-sdk-operation-name: getByVersion
```
ashkarpetin, line 16: Seems copy-pasted from `lists@{id}@{version}.yaml`

slavcoclone: @agualis could we add the check in linter to check uniqueness?

adamaltman: Yes, look at this for a hint of how to do it:

https://github.com/Redocly/openapi-cli/blob/master/packages/core/src/rules/common/operation-operationId-unique.ts

openapi/components/schemas/List.yaml
adamaltman: I think this will break at least one client (that we didn't write ourselves).

openapi/users.yaml
```diff
@@ -386,6 +386,8 @@ paths:
     $ref: './paths/lists@{id}.yaml'
   '/lists/{id}/{version}':
     $ref: './paths/lists@{id}@{version}.yaml'
+  '/lists/{id}/{version}/values':
+    $ref: './paths/lists@{id}@{version}@values.yaml'
```
ashkarpetin, line 390: It will require 2 requests to retrieve latest version of a list:

1.  `/lists` or `/lists/{id}`
2.  `/lists/{id}/{version}/values`

Retrieving latest version of a list is the most usable API.

openapi/components/schemas/List.yaml
2e3s: There should be a BC layer to output this field for older consumers.

adamaltman (changes requested):
Maybe this needs a discussion on slack or a GitHub issue I can better understand.

2e3s:
This is too low priority to continue as nobody is affected right now. The barely used API can live with the flaw.

PR #445, opened 2021-04-30
Add x-roleTemplates with lint

agualis:


root-lint/lint.js
```diff
@@ -0,0 +1,50 @@
...
+
+console.log('🤖 Checking x-roleTemplates permissions')
+
+const combinedSchema = readDistYaml();
+const roleTemplates = combinedSchema['x-roleTemplates'];
+const allOperationIds = listAllOperationIds(combinedSchema);
+roleTemplates.forEach(role => {
+    const roleId = role.id;
+    const permissions = role.permissions;
+    verifyAllPermissionsExist(roleId, permissions, allOperationIds);
```
slavcoclone, line 18: This checks that operations in templates exists, what about the check that every operation in the `paths` exists at least in one role template? This way we will assert that new operation added in API is not forgotten to be added in role templates.

agualis: It is a good idea but, are we using all the declared operations in role templates? I don't think so. 

We can create a list of excluded `operationIds` for those cases.

ghost (dismissed):

plugins/rules/valid-role-templates.js
```diff
@@ -0,0 +1,66 @@
+module.exports = ValidRoleTemplates
+
+const seenOperations = new Set();
+
+const notRegisteredOperations = [
+    // Legacy operations
+    'PostSubscriptionLegacyCancellation',
+    'PostSubscriptionPlanChange',
+    'GetInvoiceItem',
+    // Storefront does not added in `combined`
```
ashkarpetin, line 10:
```suggestion
    // Storefront is not present in `combined`
```

ghost (approved):

slavcoclone:
@Rebilly/php please re-review

slavcoclone:
@Rebilly/js please re-review

openapi/combined.yaml
```diff
@@ -682,3 +684,513 @@ paths:
     $ref: ./paths/reports@transactions.yaml
   '/subscriptions/{subscriptionId}/summary-metrics':
     $ref: './paths/subscriptions@{subscriptionId}@summary-metrics.yaml'
+x-roleTemplates:
```
adamaltman, line 687: I think we should be eliminating the `combined.yaml` file. I merged a few files in the past and made the `combined.yaml` but it's not really kept up to date.

ashkarpetin (approved):

openapi/combined.yaml
agualis: I agree about eliminating `combined.yaml ` but it is not trivial as, for example, [Recomm is using it](https://github.com/Rebilly/frontend-recomm/blob/2be96c81433fc0670fc99b652eb408f8fa3f4493/.env#L34).

openapi/combined.yaml
```diff
@@ -1170,3 +1170,27 @@ x-roleTemplates:
       - DeleteTagCustomerCollection
       - PostTagCustomer
       - DeleteTagCustomer
+# Operation IDs which present in `x-roleTemplates` but does not present in `paths` and should be ignored by linter.
+x-roleTemplatesExtraOperationIds:
+  # Legacy operations
+  - PostSubscriptionLegacyCancellation
+  - PostSubscriptionPlanChange
+  - GetInvoiceItem
+  # Storefront is not present in `combined`
```
agualis, line 1179: @slavcoclone These solution works but it's not ideal. If we would delete `combined.yaml` as @adamaltman suggests, could we somehow create an automatic merge of our 3 definitions (core, users and storefront) to be able to have features like this global operation validation?

slavcoclone: It does not make sense to remove `combined.yaml` and merge 3 specs runtime. We used to merge them in the pasdt, it was worse than have this `combined.yaml`.

agualis: > It does not make sense to remove `combined.yaml` and merge 3 specs runtime to me. We used to merge them in the past, it was worse than have this `combined.yaml`.

In the case of keeping `combined.yaml`, what about splitting `x-roleTemplates`  in 3 (core, users, storefront) that can be locally validated and globally merged when needed?

slavcoclone: > In the case of keeping combined.yaml, what about splitting x-roleTemplates in 3 

I don't know. Are all our roles uses only operations from one API description?

slavcoclone: > consistent across definitions but that should be easy

urghh, if I've understood correct you propose to have same roles in different files, with different allowed operations and merge them for generator? This sounds tricky to me.

agualis: Yep, that was my suggestion. It is a bit tricky but it would allow us to define the `permissions` in the same place where the related `operations` are located. 

The current solution is also tricky because we are referencing `storefront` permissions in a `combined.yaml` spec that does not contain storefront `operations`.

Maybe there's a third better solution. I don't know.

slavcoclone: I know no reason why storefront is not added in combined description.

PR #446, opened 2021-04-30
Check SDK operation name to be unique

2e3s:
By analogy https://github.com/Redocly/openapi-cli/blob/master/packages/core/src/rules/common/operation-operationId-unique.ts

### DIFF
--- a/plugins/custom-rules.js
+++ b/plugins/custom-rules.js
@@ -1,12 +1,12 @@
 const DescriptionPunctuation = require('./rules/description-punctuation');
-const HasSDKOperationName = require('./rules/has-sdk-operation-name');
+const HasUniqueSDKOperationName = require('./rules/has-unique-sdk-operation-name');
 const id = 'custom-rules';
 
 /** @type {import('@redocly/openapi-cli').CustomRulesConfig} */
 const rules = {
     oas3: {
         'description-punctuation': DescriptionPunctuation,
-        'has-sdk-operation-name': HasSDKOperationName,
+        'has-sdk-operation-name': HasUniqueSDKOperationName,
     },
 };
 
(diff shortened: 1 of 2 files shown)


2e3s:
`x-sdk-operation-name` should not be globally unique, its uniqueness depends on some root object name which I haven't found in API defs.

slavcoclone:
> its uniqueness depends on some root object name which I haven't found in API defs.

I guess it's related to operation tag. @agualis can you confirm?

agualis:
> > its uniqueness depends on some root object name which I haven't found in API defs.
> 
> I guess it depends on operation tag. @agualis can you confirm?

It should be unique at `"sdk resource level"` as one resource file cannot have two functions with the same name.

The [mapping](https://github.com/Rebilly/rebilly-sdk-generator/blob/bb9048c37b21b68af7d62b3517d9f5c7af73fecb/src/sdk-generation/customizations/customizations.js#L24) between `schema paths` and `resource files` is currently done in the generator project. We could discuss if it's convenient to move that mapping to our definitions. Until then, I had added [this test](https://github.com/Rebilly/rebilly-sdk-generator/blob/bb9048c37b21b68af7d62b3517d9f5c7af73fecb/test/unit/specs/sdk-generation/from-schema-to-sdk.spec.js#L589) to detect duplications as we cannot check that uniqueness in `api-definitions`.

agualis:
> > its uniqueness depends on some root object name which I haven't found in API defs.
> 
> I guess it depends on operation tag. @agualis can you confirm?

Answered in slack.

2e3s:
This is impossible to do at the moment because it needs the mapping from other repositories.

PR #447, opened 2021-05-04
Add billing portals API

ashkarpetin:
## Details

- Added billing portal Core API
- Added billing portal Storefront API


openapi/components/schemas/BillingPortals/CommonBillingPortal.yaml
```diff
@@ -0,0 +1,88 @@
...
+  - websiteId
+properties:
+  id:
+    description: The billing portal identifier string.
+    readOnly: true
+    allOf:
+      - $ref: ../ResourceId.yaml
+  websiteId:
+    description: Billing portal website ID.
+    type: string
```
slavcoclone, line 12: I think the type is already setup in `ResourceId`, isn't it?

slavcoclone, line 12:
```suggestion
```

I think the type is already setup in `ResourceId`, isn't it?

openapi/components/schemas/BillingPortals/CommonBillingPortal.yaml
```diff
@@ -0,0 +1,88 @@
...
+    allOf:
+      - $ref: ../ResourceId.yaml
+  features:
+    description: Various features that can be enabled for the billing portal.
+    type: object
+    properties:
+      authenticateWithPassword:
+        description: Allow or disable a customer ability to authenticate with a password.
```
slavcoclone, line 20:
```suggestion
        description: Enable or disable a customer ability to authenticate with a password.
```

openapi/components/schemas/BillingPortals/CommonBillingPortal.yaml
```diff
@@ -0,0 +1,88 @@
...
+            description: Primary color for the billing portal in hexadecimal format.
+            type: string
+            maxLength: 6
+            default: '0044d4'
+          secondary:
+            description: Secondary color for the billing portal in hexadecimal format.
+            type: string
+            maxLength: 6
+            default: '0044d4'
+      links:
```
slavcoclone, line 65: Just wonder if it would be better to separate the link from style? I'm not quite understand the logic behind grouping in "customization", is this object for all kind of keys that we do not know where to put to? Why they are not in root object for example?

ashkarpetin: I dont't have answser to your question, I'm doing as similar to https://user-api-docs.rebilly.com/tag/Checkout-Forms#operation/PostCheckoutForm

slavcoclone: @justnero could you help me understand this?

ashkarpetin: Personally I think it's fine to keep this in a customization object. Why not? It's a part that defines it's setting.

slavcoclone: I'm fine with either version. I just wanna understand why it was decided to group so. The "customization" sounds too generic, for example, all props from `features` are also kind of portal customization.

slavcoclone: > will be used only by frontend

It's public API, it can be used by anything and anyone.

ghost (dismissed):

colinbird (dismissed):

2e3s (dismissed):

openapi/components/schemas/BillingPortals/CommonBillingPortal.yaml
justnero: `customization` is an object containing options that doesn't change any API behaviours but still may (or may not – depending on the implementation) affect the look and feel. `features` on the other hand must be implemented by any frontend because their value expresses the exact way API would work for a given portal. Authentication options are a feature because it affects API behaviour, while Authentication button color would not affect anything.

justnero (dismissed):

openapi/components/schemas/BillingPortals/CommonBillingPortal.yaml
slavcoclone: Thank you. Maybe the description for the objects would help to understand this faster. Also maybe `style` instead of `customization` would be clearer.

justnero: Links are not style, that's why we decided to go with the more common name here.

mattsanf (dismissed):

colinbird (dismissed):

justnero (dismissed):

openapi/components/schemas/BillingPortals/CommonBillingPortal.yaml
```diff
@@ -0,0 +1,87 @@
...
+            format: url
+          termsOfService:
+            description: The url to the website terms of service.
+            type: string
+            format: url
+  createdTime:
+    description: Billing portal created time.
+    allOf:
+      - $ref: ../ServerTimestamp.yaml
+  updatedTime:
```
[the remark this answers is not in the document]
ashkarpetin: @justnero Can you please explain?

colinbird (dismissed):

colinbird (approved):

vykhrystiuk (approved):

PR #456, opened 2021-05-12
Add nationality to KYC identity documentMatches

subenson:
## Links

- [RFC](https://github.com/Rebilly/docs-product/pull/596/files)
- [Backend](https://github.com/Rebilly/backend/pull/12770)

### DIFF
--- a/openapi/components/schemas/KycDocument/IdentityMatches.yaml
+++ b/openapi/components/schemas/KycDocument/IdentityMatches.yaml
@@ -39,3 +39,7 @@ properties:
     type: boolean
     readOnly: true
     example: true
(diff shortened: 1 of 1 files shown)


adamaltman (dismissed):
Max length is 2 characters?

subenson:
We use isoAlpha2, but the max length is 3. The reason for this is that the machine readable zone displays isoAlpha3 code, so it's very likely the customer will submit this in **POST**: `/kyc-documents/{id}/matches`. We will transform this. 

I added this restriction.

adamaltman (approved):

XaosSintez (approved):

PR #470, opened 2021-05-20
Add missing report amount and currency to transaction schema

agualis:


openapi/components/schemas/Common/CommonTransaction.yaml
```diff
@@ -91,6 +91,17 @@ properties:
     readOnly: true
     allOf:
       - $ref: ../CurrencyCode.yaml
+  reportAmount:
+    description: Transaction amount converted to organization selected report currency.
+    type: number
+    x-type: Money
+    x-currency-field: requestCurrency
```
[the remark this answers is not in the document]
agualis: Normally, when using `x-type: Money` we use `currency`by default to look for the related currency property. 

But we have some special cases like this one where `currency` is not used. So we need to explicitly use `x-currency-field`to indicate the name of the currency field in that Money property.

ashkarpetin (dismissed):

openapi/components/schemas/Common/CommonTransaction.yaml
```diff
@@ -91,6 +91,17 @@ properties:
...
+    description: Transaction amount converted to organization selected report currency.
+    type: number
+    x-type: Money
+    x-currency-field: reportCurrency
+    format: double
+    readOnly: true
+  reportCurrency:
+    readOnly: true
+    allOf:
+      - $ref: ../CurrencyCode.yaml
```
ashkarpetin, line 104: This should be moved to Transaction` schema, `CommonTransaction` schema is used by Storefront, Storfront transaction doesn't have these fields.

agualis: Moved 👍

ashkarpetin (approved):

ghost (approved):

PR #515, opened 2021-06-25
App Store

justnero:
[Preview link](https://preview.redoc.ly/rebilly-users/app-store/tag/Application-owners/#operation/PostApplication)


openapi/components/schemas/Applications/Application.yaml
```diff
@@ -0,0 +1,76 @@
...
+      - $ref: ../ServerTimestamp.yaml
+  _links:
+    type: array
+    description: The links related to resource.
+    readOnly: true
+    minItems: 1
+    items:
+      anyOf:
+        - $ref: ../Links/SelfLink.yaml
+        - $ref: ../Links/ApplicationInstanceLink.yaml
```
rensftw, line 76: I think we are missing `../Links/LogoUrlLink.yaml` as an option here?

rensftw: Oh, nevermind - my bad.

rensftw:
Do we need to update `openapi/components/schemas/Files/File.yaml` to have `./openapi/components/schemas/Links/LogoUrlLink.yaml` as an option?

openapi/components/schemas/Applications/Application.yaml
rensftw: 🤔 Thinking again about this.. 
If we had this, then it saves us one BE call from Recomm

n-fasano: You're right, I added it

rensftw (approved):

VladimirBerdnik (approved):

n-fasano (approved):

PR #520, opened 2021-07-06
Add gateway account onboarding events

XaosSintez:


viceromag (dismissed):

iamdevice (dismissed):

openapi/components/schemas/EventType.yaml
```diff
@@ -9,6 +9,7 @@ enum:
   - customer-one-time-password-requested
   - dispute-created
   - experian-check-performed
+  - gateway-account-activated
```
ashkarpetin, line 12: Is this event emitted every time gatewa account status is changed to active?

XaosSintez: We also have `gateway-account-enabled` event to handle situation when gateway account was disabled and enabled again.

So, `gateway-account-activated` event is emitted only for gateways with complex onboarding

ashkarpetin: I think the event name is confusing, perhaps `gateway-account-onboarding-completed` might be better.

XaosSintez: Applied `gateway-account-onboarding-completed`

ashkarpetin (approved):

wmdmgv (approved):

iamdevice (approved):

vykhrystiuk (approved):

PR #521, opened 2021-07-09
Add bank-transfer-10 payment method

XaosSintez:


openapi/components/schemas/PaymentMethods/AlternativePaymentMethods.yaml
```diff
@@ -18,6 +18,7 @@ enum:
   - bank-transfer-7
   - bank-transfer-8
   - bank-transfer-9
+  - bank-transfer-10
```
ashkarpetin, line 21: I believe we should not add it

PR #528, opened 2021-07-14
Add `User.computedPermissions`

justnero:


ashkarpetin (dismissed):

openapi/components/schemas/User.yaml
```diff
@@ -34,6 +34,11 @@ properties:
     description: The user's permissions. See the format in example. Use wildcard `*` for full access.
     allOf:
       - $ref: ./AclPermissions.yaml
+  fullPermissions:
```
VladimirBerdnik, line 37: As for me `fullPermissions` sounds like an user has full access. I think we may use `nestedPermissions`, `inheritedPermissions` or something like this to make an accent on the meaning. Just a suggestion

VladimirBerdnik (approved):

removed-account-123 (approved):

ashkarpetin (approved):

PR #539, opened 2021-08-03
Storefront purchase redirectUrl

justnero:


openapi/components/schemas/Storefront/Purchase.yaml
```diff
@@ -69,3 +69,11 @@ properties:
     description: If set would be used to create customer account.
     writeOnly: true
     nullable: true
+  redirectUrl:
+    type: string
+    description: >-
+      The URL where the customer will be redirected after completing the transaction. It may be set to any path at the
```
ashkarpetin, line 75: > It may be set to any path at the related website.

What does it mean?

justnero: Let's say `Website.url` is `https://example.com` – `redirectUrl` could be set to any of these:

- `https://example.com`
- `https://example.com/some/path`
- `https://example.com/some/path?and=query`
- `https://example.com/some/path?and=query#and-fragment`

ashkarpetin: Perhaps it should mention that is supports templates:
```
The URL to redirect the end-user when an offsite transaction is completed. Defaults to the website's configured URL. You may use {id} or {result} as placeholders in the URL and we will replace them with the transaction's id and result accordingly.
```

ashkarpetin: > Let's say `Website.url` is `https://example.com` – `redirectUrl` could be set to any of these:
> 
> * `https://example.com`
> * `https://example.com/some/path`
> * `https://example.com/some/path?and=query`
> * `https://example.com/some/path?and=query#and-fragment`

I think it's clear enough that you provide any url. It's more confusing with it.

justnero: Adjusted description. Thank you!

ashkarpetin (approved):

mattsanf (approved):

PR #557, opened 2021-08-17
Add kountFraudSessionId

colinbird:
In order to support BlueSnap fraud prevention, we need to generate and supply the `fraudSessionId` to the Kount SDK which will be integrated into FramePay.  FramePay can add this field to the RiskMetaData so it can be consumed by the backend. The `fraudSessionId` is the required information that will be passed on as part of the backend API call to BlueSnap.

Can FramePay generate this session id?

>Q: Does the session identifier need to be unique?

> Yes, the session identifier must be unique over a thirty-day period. If there are duplicate session identifiers, device data originating from the Data Collection process may be erroneously connected to RIS order data. See Session ID Discussion for more information.

:question: Should we simply add the field as in this PR, or should it be part of a new subsection of RiskMetaData for proprietary fields like this? :question: 

https://developers.bluesnap.com/docs/fraud-prevention#section-device-data-checks
https://kount.github.io/docs/dc-web-browser/
https://kount.github.io/docs/dc-faq/

### DIFF
--- a/openapi/components/schemas/RiskMetadata.yaml
+++ b/openapi/components/schemas/RiskMetadata.yaml
@@ -137,6 +137,8 @@ properties:
     type: integer
     readOnly: true
   kountFraudSessionId:
-    description: Fraud session id as provided by Klout.
+    description: Fraud session id for Kount risk scoring.
     type: string
+    minLength: 10
+    maxLength: 32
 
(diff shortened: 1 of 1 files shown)


ashkarpetin:
@colinbird Please correctly format commit messages

PR #558, opened 2021-08-17
feat: payment instrument setups

adamaltman:
This allows for a consistent integration flow even when some gateways (for example, fictitious cryptoworld) can do nothing setup related.

Within the gateway account settings, someone can specify the setup instructions (this part is not included in this PR).

Then, they have a simple integration:

1. POST /setups
2. POST /transactions

And no... well if the payment method selected is ___ then do ____.

Ideas of setup instructions:

- do a 3DS authentication, then $1 auth + void
- do a $0 verification
- do nothing (some gateways don't support special types like `auth` and `void`)

### DIFF
--- a/openapi/components/schemas/GatewayAccount.yaml
+++ b/openapi/components/schemas/GatewayAccount.yaml
@@ -290,6 +290,42 @@ properties:
     description: Customer's payment instrument will "stick" to the gateway account for future transactions when enabled.
     type: boolean
     default: true
+  setupInstruction:
+    description: Instructions for processing a setup request.
+    type: object
+    properties:
+      continueOnFailure:
+        type: boolean
+        default: false
+        example: false
+        description: |
+          If set to continue on failure,
+          then it will proceed to execute the next step.
+          It will use the last step with a success in the result.
+      steps:
+        type: array
+        description: List of steps in the order they should execute.
+        items:
+          type: string
+          description: |
+              Many steps may ignore the amount from the request.
+              Most gateways will not support all of the types of steps.
+              In this case, check the gateway-specific definitions for available steps.
+              All gateways support `do-nothing`.
+          enum:
+              - authorize
+              - authorize-and-void
+              - gather-kyc
+              - use-sca
+              - get-billing-agreement
+              - do-nothing
+          x-enumDescriptions:
+            authorize: Will create an authorize transaction in the amount/currency of the request.
+            authorize-and-void: Will create an authorize transaction in the amount/currency of the request, followed by a void.
+            gather-kyc: Will collect proof of identity via the KYC document gatherer. Will ignore transaction amount/currency.
+            use-sca: Will use SCA (strong customer authentication) which includes 3dsv2, 3ds, and some specific wallets.
+            get-billing-agreement: Will get a billing agreement from a customer (for example, this is applicable to PayPal).
+            do-nothing: Will do nothing, and return an approved `setup` transaction. This is the default behavior.
   createdTime:
     description: Gateway Account created time.
     allOf:
--- a/openapi/core.yaml
+++ b/openapi/core.yaml
@@ -399,6 +399,8 @@ paths:
     $ref: './paths/transactions@{id}.yaml'
   /payouts:
     $ref: ./paths/payouts.yaml
+  /setups:
+      $ref: ./paths/setups.yaml
   '/transactions/{id}/query':
     $ref: './paths/transactions@{id}@query.yaml'
   '/transactions/{id}/update':
--- a/openapi/components/requestBodies/SetupRequest.yaml
+++ b/openapi/components/requestBodies/SetupRequest.yaml
@@ -1,0 +1,6 @@
+content:
+  application/json:
+    schema:
+      $ref: ../schemas/Common/CommonTransactionRequest.yaml
+description: Transaction resource.
+required: true
--- a/openapi/paths/setups.yaml
+++ b/openapi/paths/setups.yaml
@@ -11,7 +11,7 @@ post:
     Ideas of setup instructions:
       - do a 3DS authentication, then $1 auth + void
       - do a $0 verification
-      - do nothing (some gateways don't support special types like `auth` and `void`)
+      - do nothing (some gateways don't support special types like `auth` and `void`).
   requestBody:
     $ref: ../components/requestBodies/SetupRequest.yaml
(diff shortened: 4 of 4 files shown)


openapi/components/schemas/Common/CommonSetupRequest.yaml
```diff
@@ -0,0 +1,68 @@
+type: object
+required:
+  - websiteId
```
ashkarpetin, line 3: Should we have `websiteId` required, or we can take website from a customer?

openapi/components/schemas/Common/CommonSetupRequest.yaml
```diff
@@ -0,0 +1,68 @@
...
+  - customerId
+  - currency
```
ashkarpetin, line 5: Should we have `currency` required, or we can guess it somehow? Currency is needed for gateway account selection.

openapi/components/schemas/Common/CommonSetupRequest.yaml
adamaltman: It should be the same as creating a transaction.

openapi/components/schemas/Common/CommonSetupRequest.yaml
adamaltman: It should be the same as creating a transaction.

adamaltman:
@ashkarpetin if you have a different idea for this, please push to your own branch. I intentionally used the common transaction object. I'm going to force push to remove your changes.

ashkarpetin:
@adamaltman I don't think CommonTransactionRequest can be used for setups, it contains many props which should not be available when doing setups:

invoiceIds, isProcessedOutside, isMerchantInitiated, processedTime

amount should be optional for setup, I think

adamaltman:
I disagree. My entire goal is uniformity.

adamaltman:
> invoiceIds, isProcessedOutside, isMerchantInitiated, processedTime

These do make sense for setups.

`invoiceIds`: informational -- shows intent.
`isProcessedOutside`: setup can happen outside of Rebilly
`isMerchantInitiated`: some cases can be merchant initiatied.
`processedTime`: same -- especially for compatibility with `isProcessedOutside` to record if something was setup.

ashkarpetin:
@adamaltman 

if  `invoiceIds` is provided all related invoices will be marked as paid when transaction is approved

what about the requirement of amount?

adamaltman:
> If invoiceIds is provided all related invoices will be marked as paid when transaction is approved.

No, this is not true.

For example, if we pass an invoice id to  `invoiceIds` with transaction type `authorize` it won't mark that invoice as paid.

> What about the requirement of amount for setups? Is it required, but maybe used or not used depending on gateway and gateway setup instruction?

Amount is required. 0 is an allowed value. It is likely to be used.

ashkarpetin:
> For example, if we pass an invoice id to invoiceIds with transaction type authorize it won't mark that invoice as paid.

I've just checked, authorize will mark $0 invoices as paid.

openapi/components/schemas/GatewayAccount.yaml
```diff
@@ -290,6 +290,42 @@ properties:
...
+              - authorize-and-void
+              - gather-kyc
+              - use-sca
+              - get-billing-agreement
+              - do-nothing
+          x-enumDescriptions:
+            authorize: Will create an authorize transaction in the amount/currency of the request.
+            authorize-and-void: Will create an authorize transaction in the amount/currency of the request, followed by a void.
+            gather-kyc: Will collect proof of identity via the KYC document gatherer. Will ignore transaction amount/currency.
+            use-sca: Will use SCA (strong customer authentication) which includes 3dsv2, 3ds, and some specific wallets.
```
ashkarpetin, line 326: Will this create a `3ds-authentication` transaction?

openapi/components/schemas/GatewayAccount.yaml
```diff
@@ -290,6 +290,42 @@ properties:
...
+    properties:
+      continueOnFailure:
+        type: boolean
+        default: false
+        example: false
+        description: |
+          If set to continue on failure,
+          then it will proceed to execute the next step.
+          It will use the last step with a success in the result.
+      steps:
```
ashkarpetin, line 305: This will require a creating a chain of transactions if 2 or more steps are chosen?

```
authorize
authorize-and-void
``

Comment (ashkarpetin) [reply openapi/components/schemas/GatewayAccount.yaml:305]: And order matters? `authorize-and-void`, then `use-sca`,  then `gather-kyc`? This greatly complicates transaction processing implementation.

Comment (adamaltman) [reply openapi/components/schemas/GatewayAccount.yaml:305]: No, I think you misunderstand. A transaction can already be connected to multiple transactions. We'll return the transaction with type `setup` (see #562 )

File: openapi/components/schemas/GatewayAccount.yaml (hunk 3 above)
Comment (adamaltman) [reply openapi/components/schemas/GatewayAccount.yaml:326]: No, that type should be deprecated and replaced with type `setup`.

File: openapi/components/schemas/GatewayAccount.yaml [hunk 5]
```diff
@@ -290,6 +290,42 @@ properties:
...
+              In this case, check the gateway-specific definitions for available steps.
+              All gateways support `do-nothing`.
+          enum:
+              - authorize
+              - authorize-and-void
+              - gather-kyc
+              - use-sca
+              - get-billing-agreement
+              - do-nothing
+          x-enumDescriptions:
```
2e3s, line 322: This is not reflected in the UI (if it should)

openapi/core.yaml
```diff
@@ -399,6 +399,8 @@ paths:
     $ref: './paths/transactions@{id}.yaml'
   /payouts:
     $ref: ./paths/payouts.yaml
+  /setups:
+      $ref: ./paths/setups.yaml
```
2e3s, line 403: Why this approach with `/setups/` endpoint uses many steps and the other one with only `type=setup` has one step? Is it an intentional difference or just an example?

adamaltman: This was an earlier idea. This PR should be closed, and use type=setup instead.

openapi/components/schemas/GatewayAccount.yaml
adamaltman: It will be... but let's discuss in #562

PR #559, opened 2021-08-19
Remove deprecated Transaction paymentInstrument property

ashkarpetin:


openapi/components/schemas/PaymentInstructions/PaymentMethods.yaml
```diff
@@ -13,3 +13,9 @@ properties:
       Additional steps will occur via a redirect to `approvalUrl`.
     items:
       $ref: ../PaymentMethods/PaymentMethod.yaml
+  receivedBy:
```
ashkarpetin, line 16: I've remove deprecated paymentInstrument schema and moved additional properties of `cash`, `check` payment methods in the root of PaymentMethods schema. What do you think, should we move it into a separate schema `CashInstruction` and `CheckInstruction`?

adamaltman (approved):
I think this is okay.

colinbird (approved):

PR #560, opened 2021-08-19
Add ExtraData property to RiskMetadata

colinbird:
In order to support BlueSnap fraud prevention, we need to capture the `fraudSessionId` from the Kount SDK which will be integrated into FramePay. FramePay can add this field to the RiskMetaData so it can be consumed by the backend. The `fraudSessionId` is the required information that will be passed on as part of the backend API call to BlueSnap.

https://developers.bluesnap.com/docs/fraud-prevention#section-device-data-checks
https://kount.github.io/docs/dc-web-browser/
https://kount.github.io/docs/dc-faq/


viceromag (approved):

ninjasimon (approved):

ashkarpetin (approved):

vykhrystiuk (approved):

openapi/components/schemas/ExtraData.yaml
```diff
@@ -0,0 +1,10 @@
+type: object
+title: Extra data
+description: Third party data used for risk scoring.
+properties:
+  kountFraudSessionId:
```
rensftw, line 5: Perhaps`fingerprint` should be in `extraData`, since it is also collected by a third party tool.
Thoughts?

ashkarpetin: I would not to touch `fingerprint`, and definitely not in scope of this task. It's minor API change requires a lot of internal changes including BC support.

openapi/components/schemas/RiskMetadata.yaml
```diff
@@ -16,6 +16,8 @@ properties:
     $ref: ./HttpHeaders.yaml
   browserData:
     $ref: ./BrowserData.yaml
+  extraData:
```
ashkarpetin, line 19: Perhaps store `fraudSessionId` in the object root?

openapi/components/schemas/ExtraData.yaml
ashkarpetin: > since it is also collected by a third party tool.

That's not 100% true, it can be any value

openapi/components/schemas/RiskMetadata.yaml
rensftw: cc: @subenson @colinbird

PR #562, opened 2021-08-22
feat: add setupInstructions to gateway account

adamaltman:
Adds a new property to gateway accounts where the merchant can define setup instructions.

Adds a new transaction type `setup`. 

Alternative to: #558 and #561


openapi/components/schemas/Common/CommonTransaction.yaml
```diff
@@ -27,6 +27,7 @@ properties:
       - credit
       - refund
       - sale
+      - setup
```
ashkarpetin, line 30: This does not look simple. We still need to know original transaction type (auth or 3ds-authentication) for gateway integrations to work.

ashkarpetin: Boolean readOnly flag `isSetupTransaction` looks a lot easier to implement for me.

openapi/components/schemas/GatewayAccount.yaml
```diff
@@ -290,6 +290,41 @@ properties:
...
+    description: |
+      The `setupInstruction` will run a set of instructions to set up a new payment instrument when a transaction request with type `setup` is sent.
+    properties:
+      # TODO: requesting KYC is not in the current scope but is accounted for in the design.
+      # requestKyc:
+      #   # TODO: we don't need the `customerId` again.
+      #   # Should configure failure behavior (continue or end)
+      #   # Should configure sequence -- before or after transaction requests
+      #   $ref: ./KycRequest.yaml
+      transactionRequests:
```
adamaltman, line 304: Could use a better name here...

- mode

Could make `setupInstruction` a `string` (I think this is my preference).

openapi/components/schemas/Common/CommonTransaction.yaml
adamaltman: I think you misunderstand this feature.

adamaltman: What is original transaction type?

adamaltman: I think I understand your concern. It would have a child transaction type `authorize` for example.

ashkarpetin: I'm not sure I understand, we create a transaction with type `setup` which creates child transaction type `authorize`. How we would know if setup was succesful or not? We would need to iterate over each child transaction or update parent transaction status?

adamaltman: This is all described in the RFC.

adamaltman: (The parent is updated)

.redocly.yaml
```diff
@@ -27,6 +27,7 @@ lint:
     custom-rules/has-sdk-operation-name: error
 referenceDocs:
   pagination: section
+  maxDisplayedEnumValues: 10
```
adamaltman, line 30:
```suggestion
```

openapi/components/schemas/GatewayAccount.yaml
2e3s: I don't understand the reason for `transactionRequests` as well.

2e3s: Alternative: https://github.com/Rebilly/api-definitions/pull/590

adamaltman: I prefer setupInstruction as a string.

PR #567, opened 2021-08-27
Add optional transaction to KycRequest

subenson:


adamaltman:
Is this coupling necessary?

PR #568, opened 2021-08-27
feat: billing portal addons and bump offers

Weetbix:
This is a draft API specification for billing portal addon and bump offers.


openapi/components/schemas/BillingPortal/BillingPortalOffer.yaml
```diff
@@ -0,0 +1,68 @@
+title: Billing portal offer
+type: object
+required:
+  - filter
+  - addonPlans
+  - bumpPlans
+properties:
+  filter:
```
adamaltman, line 8: I'm not clear what context the filter has. Is it like a customer context? An order/subscription context? Or something else?

I think it is critical to be able to filter the offers to show. For example, let's say I have bronze, silver, gold, and diamond plans. If someone is on the gold plan, I only have the diamond plan to upsell/offer to them.

Weetbix: Hmm good question.

This is more complicated than the current checkout forms we have, as we do not filter at all, and the checkout just has a list of bump offers and add-ons which I assume are always shown.

Would we also want to filter by other properties which are not on the order context, for example customer age?

If so we might need multiple filters:
```
orderFilter: "...",
customerFilter: "...",
```

@Rebilly/php maybe someone can give some insight here

PR #575, opened 2021-09-02
Add STP Mexico gateway

ninjasimon:
Add STP Mexico gateway.


colinbird (dismissed):

openapi/components/schemas/GatewayAccountConfig/STPMexico.yaml
```diff
@@ -0,0 +1,40 @@
...
+          beneficiaryName:
+            type: string
+            description: The name of the bank account holder receiving payments.
+          bankId:
+            description: The ID of the bank receiving payments. A list of bank IDs can be found in STP Mexico integration documentation.
+            $ref: ../STPMexicoBanks.yaml
+          bankAccountNumber:
+            type: string
+            description: The account number of the bank account receiving payments.
+            format: password
```
colinbird, line 24: Does this need to be hidden?

ninjasimon: I'm not completely sure. I wasn't able to find other gateways where we collect bank account numbers in settings or credentials. This field contains a full bank account number, and, at least in my experience, after I input my account number into online services, I tend to see it at least partially masked when displayed later. Would you recommend that I remove the password formatting on it?

colinbird: I don't think it needs to be hidden

ninjasimon: Thanks for your feedback. I removed the password masking.

mikerebilly (approved):

colinbird (approved):

PR #587, opened 2021-09-09
Update error responses

slavcodev-rebilly:
This change does not work with current version or redoc, it requires the patch https://github.com/Redocly/openapi-sampler/pull/128.


slavcodev-rebilly:
I'm closing this because the https://github.com/Redocly/openapi-sampler/pull/128 seems won't be released anytime soon.

PR #608, opened 2021-09-27
Add 3DS setting for AsiaPaymentGateway

ninjasimon:
Adds a setting to enable 3DS authentication for AsiaPaymentGateway.

### DIFF
--- a/openapi/components/schemas/GatewayAccountConfig/AsiaPaymentGateway.yaml
+++ b/openapi/components/schemas/GatewayAccountConfig/AsiaPaymentGateway.yaml
@@ -19,3 +19,9 @@ allOf:
         required:
           - merchantNumber
           - secretKey
+      settings:
+        type: object
+        properties:
+          use3DS:
+            type: boolean
+            description: Use 3DS authentication.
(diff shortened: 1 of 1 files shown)


openapi/components/schemas/GatewayAccountConfig/AsiaPaymentGateway.yaml
```diff
@@ -19,3 +19,9 @@ allOf:
         required:
           - merchantNumber
           - secretKey
+      settings:
+        type: object
+        properties:
+          use3DS:
```
ashkarpetin, line 25: Why you are not using `threeDSecureServer` setting?

https://github.com/Rebilly/api-definitions/blob/ecafddadf911993e55725a9d829ee2def2bd494e/openapi/components/schemas/GatewayAccountConfig/Piastrix.yaml#L32-L33

ashkarpetin (changes requested):
~I believe it's not correct, we have a special object that instructs gateway to use 3ds or not~

It's a flag that instructs adapter to send request to a different endpoint.

ashkarpetin (approved):

vykhrystiuk (approved):

XaosSintez (approved):

PR #616, opened 2021-10-07
Add openapi file with all paths and API builder decorator

VladimirBerdnik:
This PR includes:

- openapi.yaml file with all paths, tags, webhooks, etc.
- API products configuration with required tags to include into result bundle
- A decorator that uses configurations to filter out unneeded tags from the openapi.yaml to build a bundle with only required paths and tags


VladimirBerdnik:
> What is difference between this one and https://github.com/Rebilly/api-definitions/blob/main/openapi/combined.yaml ?

this file is a replacement for users, core, reports and combined yaml files. I cannot remove these files and create new in the same PR. I need to complete my project, use new yaml file in other repos and then remove duplicates

openapi/openapi.yaml
```diff
@@ -0,0 +1,1062 @@
+openapi: 3.0.3
+info:
+  version: "2.1"
```
adamaltman, line 3: We could actually change this... not now, in the future.

plugins/products-bundler/mapping/Billing.yaml
```diff
@@ -0,0 +1,36 @@
+info:
```
slavcodev-rebilly, line 1: Maybe add these in `openapi` in some extension keyword?

plugins/products-bundler/mapping/Billing.yaml
```diff
@@ -0,0 +1,36 @@
...
+  title: "Billing API"
+x-tagGroups:
```
slavcodev-rebilly, line 3: I see `x-tagGroups` still in `openapi.yaml`, I'm confusing how that one is used and how this?

plugins/products-bundler/bundler.js
```diff
@@ -0,0 +1,152 @@
...
+const path = require('path');
+
+const availableMethods = ['get', 'put', 'post', 'delete', 'options', 'head', 'patch', 'trace'];
+
+function getProductMappingToBundle() {
+  if (!('API_BUNDLED_PRODUCT' in process.env) || !process.env.API_BUNDLED_PRODUCT) {
+    return;
+  }
+
+  const requestedProduct = process.env.API_BUNDLED_PRODUCT;
```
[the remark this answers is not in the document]
VladimirBerdnik: sure, will change the env name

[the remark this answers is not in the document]
VladimirBerdnik: for the default - I'm not sure for now. I'd like to fail the build instead. It is not used as a live script tht fail somewhere on production, it will fail on the PR's actions checks

plugins/products-bundler/mapping/Billing.yaml
slavcodev-rebilly: Sorry, I meant incorporate the data form all these "mapping" files into `openapi.yaml` using extension, something like

```
x-bundles:
  Full: ~
  Billing: ~
```

plugins/products-bundler/mapping/Billing.yaml
slavcodev-rebilly: Hmm, I haven't got then, so these would be defaults when env var is not set?

package.json
```diff
@@ -13,6 +13,11 @@
     "serve-users": "npm run start users",
     "serve-reports": "npm run start reports",
     "serve-storefront": "npm run start storefront",
-    "serve-combined": "npm run start combined"
+    "serve-combined": "npm run start combined",
+    "serve-openapi": "npm run start openapi",
+    "serve-openapi-billing": "REBILLY_API_BUNDLED_PRODUCT=Billing npm run start openapi",
```
adamaltman, line 18: What about `REBILLY_API_BUNDLE` for the env variable name? It is a little more compact.

adamaltman: Or `REBILLY_API_PRODUCT`

plugins/products-bundler/mapping/Billing.yaml
VladimirBerdnik: We may need to override more values from the `info` section, not only title. See the scope-hammered for now Storefront API product: https://github.com/Rebilly/api-definitions/pull/616/commits/6901e4dda49d87e82afb50c78ba1f841af9c97be#diff-e644157d522f9117c35053fd4233cbf14df656e94a19aabda5456f3e21ebc9fcR1-R15

plugins/products-bundler/mapping/Billing.yaml
VladimirBerdnik: yes

VladimirBerdnik: openapi will include storefront as well, I think it should not be included into Full

slavcodev-rebilly: Why? What does the "full" mean if it does not contain all API that Rebilly supports?

plugins/products-bundler/mapping/Billing.yaml
VladimirBerdnik: your approach will work, but It might be inconvenient to have everything in the single file, the description section may be long, for example, and it will be hard to navigate in single openapi file of 1k+ lines. Is there any issues with having mapping in the separate file?

plugins/products-bundler/mapping/Billing.yaml
VladimirBerdnik: Full is used to be referenced from docs website. Full means it's not only Billing API or Risk Management API

plugins/products-bundler/mapping/Billing.yaml
VladimirBerdnik: you still have to understand what `x-bundles` is and why it contains `x-tagsGroup` and other tags, similar to property names in rest of the openapi.yaml file. I think this decision must not block the PR. It is better to merge it faster to avoid possible  mess and mismatching with the `users.yaml`, `core.yaml`, `reports.yaml`, etc. There is always a better solution to implement a feature, but every improvement may be performed in further PRs

slavcodev-rebilly: And what do you suggest to do to learn more what does "bundles" mean?

BTW, I didn't block the PR tho.

plugins/products-bundler/mapping/Billing.yaml
VladimirBerdnik: It's a very specific product and I think it should be a separate bundle. It has similar paths like "Get payment instruments collection"

plugins/products-bundler/mapping/Billing.yaml
slavcodev-rebilly: > I don't think it will be much easier to understand what x-bundles mean.

Oh, I misread you previous comment. so yes you have to understand what `x-bundles`, but it's easier to notice new extension keyword in `openapi.yaml` and try to find information about it, rather than magic with CI plugins. I think for similar reason we added paths in `openapi.yaml` explicitelly, and not magically from directory `paths`.

plugins/products-bundler/mapping/Billing.yaml
slavcodev-rebilly: I'm sorry, I still can't understand why `Storefront API` is more specific than, let say `Risk Management API`, and can't understand what is the real purpose of `Full` if it does not contain all API that Rebilly support.

> Full is used to be referenced from docs website

I thought we create bundles by product: Billing, Payments, exactly for website docs, what do you mean more exactly by "will be referenced"? Could we reference the product description instead?

ghost (approved):
> It is better to merge it faster to avoid possible mess and mismatching with the users.yaml, core.yaml, reports.yaml, etc

Although, I'm really have not got what issues with "users.yaml, core.yaml, reports.yaml, etc", I guess they are going to be removed when this one is released.

I also, has a few concerns I mentioned in comments.

But, I think PR can be merged as it is for now, since it contains a big work, good as a single step forward.

plugins/products-bundler/mapping/Billing.yaml
slavcodev-rebilly: > For the blocking - any unresolved conversation is a blocker for the rest of the team :)

This sounds wrong to me, anyway I do my best to unblock it ;) even my concerns are not solved.

slavcodev-rebilly: > BTW, do you think we may use something like this:

Yes. That's my idea. Maybe with different wording (e.g. "mapping" => "extend/merge/or something else").

ashkarpetin (approved):

PR #630, opened 2021-10-16
Products in operations

VladimirBerdnik:
This PR will allow to include specific operations into products bundles.
With this feature we can remove `users.yaml`, `core.yaml`, `combined.yaml`, `reports.yaml` and `storefront.yaml` but still build  these bundles with the help of `x-products` property and the products bundler plugin


VladimirBerdnik:
devs please review

2e3s (approved):

VladimirBerdnik:
devs please review

ashkarpetin (approved):

adamaltman (approved):
Overall, I think this should be changed. The x-products seems redundant to me. Also, I'm not clear on the complete list... is it Core, Full, Risk Management, Users, Payments, Billing? (With that being said, I don't think this is harmful and we need to make the switch sooner rather than later.)

adamaltman:
You already have a conflict to resolve. 🎃

PR #634, opened 2021-10-20
KYC document type identity-face-proof

justnero:


adamaltman:
Is this still relevant? @swilczynski ?

swilczynski:
@adamaltman it is not, we decided to reuse and extend the existing identity-proof

PR #635, opened 2021-10-20
Add KYC settings

subenson:


openapi/code_samples/JavaScript/settings@kyc/get.js
```diff
@@ -0,0 +1,2 @@
+const kycSettings = await api.kycSettings.get();
+console.log(kycSettings.requireDateOfBirth);
```
p3yman, line 2: Is this for debugging or should be removed?

openapi/components/schemas/Common/CommonKycDocument.yaml
```diff
@@ -56,6 +56,9 @@ properties:
     description: Processing date/time.
     allOf:
       - $ref: ../ServerTimestamp.yaml
+  settings:
+    readOnly: true
+    $ref: ../KycSettings.yaml
```
slavcodev-rebilly, line 61: This is not correct schema according OAS 3.0.3 we use. if `$ref` keyword is used, it it should single in that level. You can use this instead if composition is needed
```yaml
readOnly: true
allOf:
    - $ref: ../KycSettings.yaml
```

slavcodev-rebilly: BTW, if this settings are read-only, how we can set them?

subenson: For now it's set manually for merchants. We hammered the scope of our project.

ghost (approved):

PR #642, opened 2021-10-26
Add servers options with organizationId server template variable

adamaltman:
Adds server template variables.


openapi/reports.yaml
```diff
@@ -58,6 +58,18 @@ security:
   - SecretApiKey: []
 servers:
   - url: 'https://api.rebilly.com/experimental'
```
ashkarpetin, line 60: It is not removed, is it expected?

ashkarpetin (dismissed):

openapi/reports.yaml
ashkarpetin, line 60: No sandbox url?

justnero: @ashkarpetin Good point, added

ashkarpetin (approved):

akozhevnikov-rebilly (approved):

PR #643, opened 2021-10-26
Support Apple Pay domain registration

colinbird:
https://developer.apple.com/documentation/applepaywebmerchantregistrationapi


adamaltman (dismissed):

openapi/components/schemas/ApplePayDomainRegistration.yaml
```diff
@@ -0,0 +1,20 @@
+description: |
+  First you will need to download Rebilly's Apple Pay merchant domain verification file and
+  host it on your domain.  https://{DOMAIN}/.well-known/apple-developer-merchantid-domain-association.
```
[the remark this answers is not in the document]
colinbird: @adamaltman I tried but that location doesn't work https://www.rebilly.com/.well-known/apple-developer-merchantid-domain-association.  We can troubleshoot this or just host it at something like https://www.rebilly.com/files/apple-pay/apple-developer-merchantid-domain-association

adamaltman (dismissed):

ashkarpetin (dismissed):

openapi/paths/digital-wallets@onboarding@apple-pay.yaml
```diff
@@ -0,0 +1,35 @@
...
+        schema:
+          $ref: ../components/schemas/DigitalWalletOnboardingApplePay.yaml
+    description: Apple pay domain registration onboarding request.
+    required: true
+  security:
+    - SecretApiKey: []
+    - JWT: []
+  responses:
+    '201':
+      description: Apple Pay onboarding was successful.
```
colinbird, line 25:
```suggestion
      description: Apple Pay onboarding attempt proceeded without error.
```

openapi/paths/digital-wallets@onboarding@apple-pay.yaml
```diff
@@ -0,0 +1,41 @@
...
+    - JWT: []
+  responses:
+    '200':
+      description: The action executed without error.
+      content:
+        application/json:
+          schema:
+            $ref: ../components/schemas/DigitalWalletOnboardingApplePay.yaml
+    '201':
+      description: Apple Pay onboarding attempt proceeded without error.
```
colinbird, line 31:
```suggestion
      description: Domain was successfully registered with Apple Pay.
```

mikerebilly (approved):
I agree with using 2xx if we use the response body as an extra layer to retrieve response/state from upstream API. 2xx to this API was a successful call, but upstream API may have an error, so technically 2xx makes sense in our API response.

colinbird:
> I agree with using 2xx if we use the response body as an extra layer to retrieve response/state from upstream API. 2xx to this API was a successful call, but upstream API may have an error, so technically 2xx makes sense in our API response.

Thanks @mikerebilly I was debating that

mikerebilly:
My second thought is a 502 response, since it's an upstream issue and in this case our API call is a proxy.

akozhevnikov-rebilly (approved):

PR #646, opened 2021-10-27
PayPal in RTP and enhanced payment token

adamaltman:
For corresponding alternative RFC.


justnero (commented):
If multiple merchantIds are available – multiple RTP items with different features would be generated (as in Plaid)

adamaltman:
Here is a sample comment [preview](#)

slavcodev-rebilly:
@Rebilly/js please review

ashkarpetin (approved):

openapi/components/schemas/ReadyToPay/ReadyToPayMethods.yaml
```diff
@@ -1,9 +1,10 @@
 type: array
 items:
-  oneOf:
+  anyOf:
```
slavcodev-rebilly, line 3: I think this is incorrect, but it was in initial proposal, so I left it as is.

ghost (approved):

PR #649, opened 2021-11-02
Require a billing address for a payment card

2e3s:
This is already the current behaviour. Other payment methods also require it.


2e3s:
It seems possible now to omit the billing address for some cases.

PR #651, opened 2021-11-03
Add token to StorefrontPostPaymentInstrument as customer JWT for further requests

wmdmgv:
Add token to StorefrontPostPaymentInstrument as customer JWT for further requests


mattsanf:
StorefrontPostPaymentInstrument still require a CustomerJwt to be called in the first place? The docs say's a CustomerJwt is required, if so token will always be null

Screen Shot 2021-11-03 at 2 55 24 PM

mattsanf (approved):

ashkarpetin (approved):

subenson (approved):

PR #660, opened 2021-11-09
Add shipping and notes to orders

subenson:


openapi/components/schemas/Common/CommonInvoice.yaml
```diff
@@ -55,7 +55,7 @@ properties:
     format: double
     readOnly: true
   shipping:
-    $ref: ../Invoices/InvoiceShipping.yaml
+    $ref: ../Invoices/Shipping.yaml
```
ashkarpetin, line 58: Should be moved to root? Since it used by Invoice and Orders?

subenson: Good idea.

ninjasimon (approved):

ashkarpetin (approved):

PR #663, opened 2021-11-11
Adjust organization prefix for storefront and experimental APIs

justnero:


ghost (approved):
It's sad that OAS does not support $ref in `servers`. Maybe would be clearer. Maybe we can add plugin to Redoc do avoid this duplicate code.

ashkarpetin (approved):

PR #688, opened 2021-12-20
Support custom Apple Pay merchant id in Gateway Accounts and RTP

colinbird:


openapi/components/schemas/PaymentTokens/DigitalWalletToken.yaml
```diff
@@ -59,6 +59,8 @@ properties:
         readOnly: true
         description: Payment Card expiration year.
         type: integer
+      applePayMerchantId:
```
colinbird, line 62: This should be in another file ApplePayDigitalWalletToken.yaml

openapi/components/schemas/PaymentTokens/DigitalWalletValidation/ApplePayValidation.yaml
```diff
@@ -11,7 +11,10 @@ allOf:
         description: The validation request.
         type: object
         properties:
-          validationURL:
```
adamaltman, line 14: Breaking change?

colinbird: The backend accepts both.  FramePay is currently sending `validationUrl`

PR #691, opened 2021-12-21
Prepare multiple openapi entrypoints to replace with a single one

VladimirBerdnik:


adamaltman (dismissed):
Redocly is changing the `.redocly.yaml` config format. This will enable some optimization later on.

plugins/products-bundler/mapping/Core.yaml
```diff
@@ -34,6 +37,10 @@ x-tagGroups:
       - Custom Fields
       - Files
       - Search
+      - Gateway Accounts
```
ashkarpetin, line 40: Should not be in a core

VladimirBerdnik: It is required to include x-webhooks with this tag. Only webhooks are included without operations

plugins/products-bundler/mapping/Core.yaml
```diff
@@ -34,6 +37,10 @@ x-tagGroups:
...
+      - Application owners
+      - Application users
+      - Balance transactions
```
ashkarpetin, line 43: All of these are in the Users

VladimirBerdnik: These are required to include x-webhooks, no operations will be added
image

ashkarpetin: These webhooks should be in users

VladimirBerdnik: I thought all webhooks are located in Core

VladimirBerdnik: I see 5 webhooks in Users, maybe they were added by mistake

VladimirBerdnik: Adjusted tags and webhooks, thank you!

openapi/core.yaml
```diff
@@ -197,12 +206,15 @@ x-tagGroups:
   - name: Payments
     tags:
       - Transactions
+      - Transactions Timeline
```
ashkarpetin, line 209: Can we move timelines at the end of the list until we will be able to move them to Users?

VladimirBerdnik: Moved! :smile:

adamaltman (approved):

ashkarpetin (approved):

PR #694, opened 2021-12-22
Add Certificates API

colinbird:


openapi/components/schemas/Certificate.yaml
```diff
@@ -0,0 +1,18 @@
+type: object
+properties:  
+  id:
+    description: The session identifier string.
+    readOnly: true
+    allOf:
+      - $ref: ./ResourceId.yaml
+  pem:
+    description: The signed certificate in pem format.
```
VladimirBerdnik, line 9: should it be write-only?

openapi/components/schemas/Certificate.yaml
```diff
@@ -0,0 +1,22 @@
+type: object
+properties:
+  id:
+    description: The session identifier string.
```
VladimirBerdnik, line 4:
```suggestion
    description: The certificate identifier string.
```

openapi/components/schemas/Certificate.yaml
colinbird: I think it can be returned to the merchant

PR #697, opened 2021-12-27
Add void subscription endpoint

swilczynski:
PHP SDK update for void method will be in the separate PR

### DIFF
--- a/openapi/paths/subscriptions@{id}@void.yaml
+++ b/openapi/paths/subscriptions@{id}@void.yaml
@@ -15,7 +15,7 @@ post:
     Void a pending order with predefined identifier string.
   responses:
     '201':
-      description: Order was voided.
+      description: Order was voided successfully.
       content:
         application/json:
           schema:
@@ -24,8 +24,10 @@ post:
       $ref: ../components/responses/Unauthorized.yaml
     '403':
       $ref: ../components/responses/Forbidden.yaml
-    '422':
-      $ref: ../components/responses/ValidationError.yaml
+    '404':
+      $ref: ../components/responses/NotFound.yaml
+    '409':
+      $ref: ../components/responses/Conflict.yaml
   x-codeSamples:
     - lang: PHP
       source:
--- a/openapi/paths/subscriptions@{id}@void.yaml
+++ b/openapi/paths/subscriptions@{id}@void.yaml
@@ -1,0 +1,35 @@
+parameters:
+  - $ref: ../components/parameters/resourceId.yaml
+  - $ref: ../components/parameters/organizationId.yaml
+post:
+  x-products:
+    - Full
+    - Billing
+    - Core
+  tags:
+    - Orders
+  summary: Void an order
+  operationId: PostSubscriptionVoid
+  x-sdk-operation-name: void
+  description: >
+    Void a pending order with predefined identifier string.
+  responses:
+    '201':
+      description: Order was voided.
+      content:
+        application/json:
+          schema:
+            $ref: ../components/schemas/Subscription.yaml
+    '401':
+      $ref: ../components/responses/Unauthorized.yaml
+    '403':
+      $ref: ../components/responses/Forbidden.yaml
+    '422':
+      $ref: ../components/responses/ValidationError.yaml
+  x-codeSamples:
+    - lang: PHP
+      source:
+        $ref: '../code_samples/PHP/subscriptions@{id}@void/post.php'
+    - lang: JavaScript
+      source:
+        $ref: '../code_samples/JavaScript/subscriptions@{id}@void/post.js'
--- a/openapi/paths/subscriptions@{id}@void.yaml
+++ b/openapi/paths/subscriptions@{id}@void.yaml
@@ -14,8 +14,12 @@ post:
   description: >
     Void a pending order with predefined identifier string.
   responses:
-    '204':
+    '201':
       description: Order was voided.
+      content:
+        application/json:
+          schema:
+            $ref: ../components/schemas/Subscription.yaml
     '401':
       $ref: ../components/responses/Unauthorized.yaml
     '403':
--- a/openapi/paths/subscriptions@{id}@void.yaml
+++ b/openapi/paths/subscriptions@{id}@void.yaml
@@ -14,12 +14,8 @@ post:
   description: >
     Void a pending order with predefined identifier string.
   responses:
-    '201':
+    '204':
       description: Order was voided.
-      content:
-        application/json:
-          schema:
-            $ref: ../components/schemas/Subscription.yaml
     '401':
       $ref: ../components/responses/Unauthorized.yaml
     '403':
--- a/openapi/code_samples/JavaScript/subscriptions@{id}@void/post.js
+++ b/openapi/code_samples/JavaScript/subscriptions@{id}@void/post.js
@@ -1,0 +1,1 @@
+const voidedSubscription = await api.subscriptions.void({id: 'my-subscription-id'});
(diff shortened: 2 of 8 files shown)


swilczynski:
PR for the new endpoint #15107

openapi/paths/subscriptions@{id}@void.yaml
```diff
@@ -0,0 +1,31 @@
...
+  tags:
+    - Orders
+  summary: Void an order
+  operationId: PostSubscriptionVoid
+  x-sdk-operation-name: void
+  description: >
+    Void a pending order with predefined identifier string.
+  responses:
+    '204':
+      description: Order was voided.
```
ashkarpetin, line 18: Should return 201, if voided:

https://github.com/Rebilly/api-definitions/blob/770348c7598acfc5cbea5cd4c0571051ecf4962f/openapi/paths/invoices%40%7Bid%7D%40void.yaml#L17-L22

openapi/paths/subscriptions@{id}@void.yaml
```diff
@@ -0,0 +1,31 @@
...
+    - Core
+  tags:
+    - Orders
+  summary: Void an order
+  operationId: PostSubscriptionVoid
+  x-sdk-operation-name: void
+  description: >
+    Void a pending order with predefined identifier string.
+  responses:
+    '204':
```
slavcodev-rebilly, line 17: The `POST` method is not idempotent method, what does 2 sequence requests mean for that order? What does response `204` means for client in such cases?

openapi/paths/subscriptions@{id}@void.yaml
creocoder: 201 mean created, what exactly created there?

openapi/paths/subscriptions@{id}@void.yaml
creocoder: > is not idempotent method

Yes, not idempotent, that mean there are no any requirements like for idempotent methods. `204` in that case mean operation was sucessful.

slavcodev-rebilly: You have not answered to my question though. What does send request mean for 

1) domain? second void? nothing just accept and ignore? anything else?
2) for client? what does mean the second `204`?

swilczynski: Should it be 200 or 204? 200 would return voided subscription and 204 would not return anything, just an status that operation was successful

openapi/paths/subscriptions@{id}@void.yaml
slavcodev-rebilly: > what exactly created there

subscription void?

creocoder: 201 is pretty clean about what exactly should be created, its about new resource with new identifier. This actually 200.

openapi/paths/subscriptions@{id}@void.yaml
creocoder: I think 200 make sense there.

openapi/paths/subscriptions@{id}@void.yaml
slavcodev-rebilly: In the REST there is no such thing like "identifier", only resource and its URL.

openapi/paths/subscriptions@{id}@void.yaml
slavcodev-rebilly: The changing status to "200" still does not answer the question, does it?

1) what does second request to void the same order will do?
2) what does second "OK" (200) means for client?

openapi/paths/subscriptions@{id}@void.yaml
creocoder: https://pradeeploganathan.com/rest/identifying-resources-and-designing-representations/

creocoder: REST resource identifier is pretty common term.

openapi/paths/subscriptions@{id}@void.yaml
creocoder: I actually think it should be PUT with 200 to answer both.

swilczynski: Second request to void the same order is going to return "409" as it has already been voided

openapi/paths/subscriptions@{id}@void.yaml
slavcodev-rebilly: Alright, I misunderstood you sentences, I thought you're talking about resource field like "id" (or another field name).
However, looking at post you referred you mean the same what I say URI.

However you also said "Void isn't about creation, it about changing state." which is confusing, you use specific URL, which clearly locates a resource.

openapi/paths/subscriptions@{id}@void.yaml
creocoder: I think I'll to change my proposal to PUT + 204. GitHub is agree with me: https://docs.github.com/en/rest/reference/gists#star-a-gist

200 is just wasting bandwidth for this action + CPU and DB resources.

openapi/paths/subscriptions@{id}@void.yaml
creocoder: I think changing it to PUT + 204 will solve existent issues.

ashkarpetin: We have a plenty of resources that do POST+201, how are you going to fix that, otherwise, its one more inconsistensy in our API.

openapi/paths/subscriptions@{id}@void.yaml
ashkarpetin: > 200 is just wasting bandwidth for this action + CPU and DB resources.

I don't this viable argument, it will be rarely used endpoint.

swilczynski:
I think that despite all of the valid points that @creocoder  made, there is one fact that I think is crucial at this moment that is the consistency. I would rather adjust one endpoint to match multiple, not ideal, but working endpoints, than create one endpoint that is different and stands out of the rest, but matches proper REST principles better. The fact that Github does PUT + 204 is perfectly valid, the excess bandwidth, even though @ashkarpetin thinks is not critical, I still think could be considered valid just as a good practice. One endpoint may not generate large bandwidth usage but many of them in a high traffic environment will.

For now, the simplest and most consistent way in my opinion would be to adjust this endpoint to be matching what we currently have and further discuss REST actions in the brainstorming Slack channel as the concept seems to be valid and perhaps can lead to creating an RFC to migrate all of the action endpoints to PUT + 204 and update the docs and SDKs.

openapi/paths/subscriptions@{id}@void.yaml
swilczynski: Replied in the PR comment

openapi/paths/subscriptions@{id}@void.yaml
swilczynski: Replied in the PR comment

openapi/paths/subscriptions@{id}@void.yaml
```diff
@@ -0,0 +1,35 @@
...
+    - Billing
+    - Core
+  tags:
+    - Orders
+  summary: Void an order
+  operationId: PostSubscriptionVoid
+  x-sdk-operation-name: void
+  description: >
+    Void a pending order with predefined identifier string.
+  responses:
```
slavcodev-rebilly, line 16: Should we mention the `409` response for cases when request comes for already voided order?

swilczynski: Yes, thanks for noticing, I still get lost in things.

ashkarpetin (approved):

creocoder (approved):

PR #719, opened 2022-02-02
Make Invoice, PaymentCard, Subscription reminder number nullable

ashkarpetin:


justnero (approved):

openapi/components/schemas/Invoices/Invoice.yaml
```diff
@@ -77,6 +77,7 @@ allOf:
           - $ref: ../ServerTimestamp.yaml
       dueReminderNumber:
         description: Number of past due reminder events triggered.
+        nullable: true
```
ashkarpetin, line 80: I think there should an object `dueReminder` which contains number and scheduledTime, but I'll leave it as it

akozhevnikov-rebilly (approved):

PR #721, opened 2022-02-04
docs: improve kyc request docs

adamaltman:
- Also fixes some mistakes in tags (case mismatch)
- Re-arranges the order of some of the APIs so they display more important first
- Remove some timeline APIs from the core docs (not really core)

### DIFF
--- a/openapi/components/schemas/Common/CommonKycRequest.yaml
+++ b/openapi/components/schemas/Common/CommonKycRequest.yaml
@@ -43,7 +43,15 @@ properties:
       - fulfilled
       - expired
   redirectUrl:
-    description: The URL to redirect the customer when an upload is completed.
+    description: |
+      The URL to redirect the customer when an upload is completed.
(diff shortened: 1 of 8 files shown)


openapi/components/schemas/Common/CommonKycRequest.yaml
```diff
@@ -43,7 +43,15 @@ properties:
...
+      When the customer is redirected, Rebilly will append an `info` query parameter which may have one of the following values (example: `https://example.com?info=success`):
+
+        - **back** - The customer clicked the `back to website` link;
+        - **token_expired** - The customer's token expired;
+        - **success** - The customer uploaded docs that were successfully analyzed;
+        - **manual** - The customer uploaded docs that will require manual review because the analyzer rejected them or couldn't process them;
+        - **partial** - The customer had success with some doc type but not with other (for example, successful proof of address but skipped proof of id).
     type: string
     format: uri
   expirationTime:
```
adamaltman, line 57: @justnero I think this would be better if we explain the default behavior.

removed-account-123 (approved):

ninjasimon (approved):

PR #722, opened 2022-02-04
docs: poc remove deprecated servers and organization-id header from docs

adamaltman:
Removing the deprecated organization-id header and servers from the docs.

First, existing integrations should migrate to the new format even though we aren't discontinuing support. 

Second, the docs look a lot worse as they are now:

Third, there is chance new integrators will opt to use the deprecated options.

### DIFF
--- a/openapi/paths/api-keys.yaml
+++ b/openapi/paths/api-keys.yaml
@@ -1,5 +1,4 @@
-parameters:
-  - $ref: ../components/parameters/organizationId.yaml
+
 get:
   x-products:
     - Full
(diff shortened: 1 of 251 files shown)


openapi/paths/api-keys.yaml
```diff
@@ -1,5 +1,4 @@
-parameters:
-  - $ref: ../components/parameters/organizationId.yaml
+
```
XaosSintez, line 1: Is there a reason for empty first line?

adamaltman: This is a really quick find/replace in IDE... this is a POC. I would ask someone to do this with more serious eyes if we agree this is a good idea. (One of our custom plugins fails too)

ashkarpetin:
Absorbed by https://github.com/Rebilly/api-definitions/pull/745, https://github.com/Rebilly/api-definitions/pull/746

PR #727, opened 2022-02-07
Deprecate invoiceTimeShift

justnero:


akozhevnikov-rebilly (approved):

openapi/components/schemas/Common/CommonSubscriptionOrder.yaml
```diff
@@ -57,8 +57,31 @@ allOf:
           this property allows to subscribe to different plans.
         nullable: true
         example: null
+        deprecated: true
```
nanobash, line 60: Does it mean that `invoiceTimeShift` should be replaced with `invoiceIssueTimeShift`, so after some time we should delete deprecated `invoiceTimeShift` endpoint?

justnero: Yes

nanobash: Got it, thanks

nanobash (approved):

PR #743, opened 2022-02-15
Use single entrypoint

VladimirBerdnik:
This PR will remove multiple openapi files in favor of single `openapi.yaml` entrypoint.
All current references already use single entrypoint after [recent PR](https://github.com/Rebilly/api-definitions/pull/691). This PR is a final step AKA cleanup


VladimirBerdnik:
php please review

ashkarpetin (approved):

agualis (approved):

agualis (changes requested):
This change will break the js-sdk-generator. We need to understand how to fix it to be compatible with this.

VladimirBerdnik:
> This change will break the js-sdk-generator. We need to understand how to fix it to be compatible with this.

@agualis , could you please explain what exactly is broken?

agualis (approved):

agualis:
> > This change will break the js-sdk-generator. We need to understand how to fix it to be compatible with this.
> 
> @agualis , could you please explain what exactly is broken?

I re-approved the PR as [this change](https://github.com/Rebilly/rebilly-js-sdk/pull/372/files) would be  enough to fix the generator for now.

PR #788, opened 2022-03-10
Add digitalWallet to PaymentCard

colinbird:
Exposing this so we can identify payment cards that have come from a digital wallet


dara123 (dismissed):

openapi/components/schemas/PaymentCards/PaymentCard.yaml
```diff
@@ -22,6 +22,13 @@ allOf:
           type: string
         example:
           gatewayTransactionId: GAT123
+      digitalWallet:
```
colinbird, line 25:
```suggestion
      digitalWalletType:
```

?

ninjasimon (dismissed):

openapi/components/schemas/PaymentCards/PaymentCard.yaml
jvc9109: I think `digitalWallet` is more suitable given the internal name on the API source code. Should be nullable though

jvc9109 (approved):

mikerebilly (approved):

PR #785, opened 2022-03-10
fix: discriminator must be used with oneOf, anyOf, allOf

adamaltman:

PR #789, opened 2022-03-10
Improve definitions with discriminator

VladimirBerdnik:
A follow-up for https://github.com/Rebilly/api-definitions/pull/785


VladimirBerdnik:
devs please review

openapi/components/schemas/GatewayAccount.yaml
```diff
@@ -347,3 +347,171 @@ properties:
         - $ref: ./Links/OnBoardingUrlLink.yaml
         - $ref: ./Links/DynamicIpnLink.yaml
         - $ref: ./Links/StaticIpnLink.yaml
+oneOf:
```
adamaltman, line 350: I think this is incorrect. They are using allOf so they are valid. And it must be anyOf here, because many of them have the same schema.

adamaltman (changes requested):

openapi/components/schemas/GatewayAccount.yaml
VladimirBerdnik, line 350:
```suggestion
anyOf:
```

VladimirBerdnik: Site previews started to crash after this change. I don't see the reason, it only says it is crashed

openapi/components/schemas/Timelines/TimelineAction.yaml
```diff
@@ -15,3 +15,8 @@ properties:
       - redemption-cancel
       - ruleset-restore
       - show-edd-search-logs
+oneOf:
+  - $ref: ./Actions/redemption-cancel.yaml
+  - $ref: ./Actions/resend-email.yaml
+  - $ref: ./Actions/ruleset-restore.yaml
+  - $ref: ./Actions/schow-edd-search-logs.yaml
```
adamaltman, line 22:
```suggestion
  - $ref: ./Actions/show-edd-search-logs.yaml
```

VladimirBerdnik:
Closing due to age

PR #793, opened 2022-03-11
Add KYC related metrics

2e3s:


openapi/components/schemas/DashboardResponse.yaml
```diff
@@ -33,6 +33,9 @@ items:
         - newSubscriptionsCount
         - upgradesCount
         - downgradesCount
+        - kycReviewTime
+        - kycRejectionRate
```
adamaltman, line 37: I don't think we need this one.

2e3s:
To be added separately

PR #798, opened 2022-03-18
Fix coupon-only restrictions

mikerebilly:
Some of the restrictions were listed for both Coupon and Redemptions. This fixes the ones that are not applicable to Redemptions as additional restrictions, only for Coupon creating or updating.


openapi/code_samples/PHP/coupons@{id}/put.php
```diff
@@ -1,24 +1,24 @@
-$couponForm = new Rebilly\Entities\Coupons\Coupon();
+$couponForm = new Rebilly\Entities\Coupons\Coupon([
+        'issuedTime' => '2022-01-01T00:00:00-04:00'
```
ashkarpetin, line 2:
```suggestion
    'issuedTime' => '2022-01-01T00:00:00-04:00',
```

ashkarpetin (approved):

colinbird (approved):

PR #836, opened 2022-04-17
test: add assertions

adamaltman:
Adds the opening example assertions:

https://redocly.com/docs/cli/resources/rules/assertions/#examples

Hit an [issue](https://github.com/Redocly/openapi-cli/issues/645).


adamaltman:
I added these in a different way already. This is outdated. Closing.

PR #864, opened 2022-05-16
Add revision to KycRequest

ashkarpetin:


dara123 (approved):

openapi/components/schemas/KycRequest.yaml
```diff
@@ -19,6 +19,13 @@ allOf:
         minimum: 1
         maximum: 2
         default: 2
+      revision:
+        description: >
+          The number of times the KYC request data has been modified.
+          The revision is useful when analyzing webhook data to determine if the
+          change takes precedence over the current representation.
+        type: integer
+        readOnly: true
```
colinbird, line 28: minValue useful ?

ashkarpetin: It's a readonly property so I don't think we need it. Also we don't have it for other resources.

ninjasimon (approved):

colinbird (approved):

PR #928, opened 2022-07-19
Add title to inline objects for php sdk generation

jvc9109:


openapi/paths/customers@{id}@edd-score.yaml
```diff
@@ -37,6 +37,7 @@ patch:
       application/json:
         schema:
           type: object
+          title: PatchCustomerEddScoreRequest
```
VladimirBerdnik, line 40: Isn't it the same as `operationId` at least it looks similar in other files in this PR?

jvc9109: `operationId` is used in a customized generator in order to generate the proper php method. 
In this case `title` is used by a default feature of the generator we extend.

openapi/paths/tags@{tag}@kyc-documents.yaml
```diff
@@ -55,6 +56,7 @@ delete:
       application/json:
         schema:
           type: object
+          title: DeleteTagKycDocumentCollectionRequest
```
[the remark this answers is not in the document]
jvc9109: All that are `Request` should have the suffix. I will look for those that does not include it for correct them. Thanks for noticing!

openapi/paths/customers@{id}@edd-score.yaml
VladimirBerdnik: Do you think it will work if we automate this by using `{operationId} + "Request"` and `{operationId} + "Response"` instead of manual editing?

VladimirBerdnik: I mean we can write a plugin for our API definition that will put titles automatically into result bundle so we don't need to write titles manually for existing and future requests and responses. We won't touch the generator code

jvc9109: Oh I am sorry, now I understand what you mean. It is a good idea. Thanks for your comment!

VladimirBerdnik: On the other hand if there are only a couple of titles that are listed in this PR we can always automate it later

VladimirBerdnik (approved):

justnero (approved):

PR #947, opened 2022-08-01
refactor PatchTransactionRequest to be generated with php-sdk

jvc9109:
small change needed for php-sdk generation


openapi/paths/transactions@{id}.yaml
```diff
@@ -47,14 +47,18 @@ patch:
...
+        schema:
+          $ref: ../components/requestBodies/PatchTransactionRequest.yaml
   responses:
     '200':
       description: Transaction was updated successfully.
       content:
         application/json:
           schema:
             $ref: ../components/schemas/Transactions/Transaction.yaml
+            required: true
```
justnero, line 61: Why is it here?

openapi/components/requestBodies/PatchTransactionRequest.yaml
```diff
@@ -1,9 +1,5 @@
...
-  application/json:
-    schema:
-      type: object
-      properties:
-        customFields:
-          $ref: ../schemas/ResourceCustomFields.yaml
+type: object
+properties:
+  customFields:
+    $ref: ../schemas/ResourceCustomFields.yaml
```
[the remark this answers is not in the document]
ashkarpetin: That requestBody indeed, so this change looks strange to me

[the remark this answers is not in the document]
jvc9109: It is a requestBody, but unlike the others, this request body refers to a `resource` rather than to a `schema`. This conflicts with the php-sdk generation since the `ResourceCustomFields` is not generated as model but when it is used it is an `array`. 

I can explore a different approach within the `PhpGenerator` instead of modify the `requestBody`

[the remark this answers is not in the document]
ashkarpetin: I am fine with separate schema, but the change is incorrect, since PatchTransactionRequest is not a requestBody anymore

openapi/components/requestBodies/PatchTransactionRequest.yaml
```diff
@@ -1,9 +1,6 @@
...
-  application/json:
-    schema:
-      type: object
-      properties:
-        customFields:
-          $ref: ../schemas/ResourceCustomFields.yaml
+content: 
+  application/json: 
+    schema: 
+      $ref: ../schemas/Transactions/PatchTransactionRequest.yaml
```
justnero, line 4: Remove trailing spaces
```suggestion
content:
  application/json:
    schema:
      $ref: ../schemas/Transactions/PatchTransactionRequest.yaml
```

justnero (dismissed):

ashkarpetin (dismissed):

adamaltman:
Is this because you need the schemas to be separated from the `requestBody`? If so, we have quite a few more like this...

openapi/components/schemas/Transactions/PatchTransactionRequest.yaml
```diff
@@ -0,0 +1,4 @@
+type: object
+properties:
```
adamaltman, line 2: I prefer to add a title when making a standalone schema.

openapi/components/requestBodies/PatchTransactionRequest.yaml
```diff
@@ -1,9 +1,6 @@
 content:
   application/json:
     schema:
-      type: object
-      properties:
-        customFields:
-          $ref: ../schemas/ResourceCustomFields.yaml
+      $ref: ../schemas/Transactions/PatchTransactionRequest.yaml
```
adamaltman, line 4: We can have a lint rule to enforce this -- I tried it but disabled it because there are a lot of violations (especially around arrays/collections).

justnero: That would help a lot in further generation steps and for the overall uniformity of our API.

justnero (approved):

ashkarpetin (approved):

jvc9109:
@adamaltman It is true that there is a bunch more of examples, but looks like the generator for core-api only has problems with this particular case. For the remaining cases  probably the way to go will be to add a lint rule and fix them all when expanding support of the sdk generator for the other APIs.

adamaltman:
@jvc9109 note that won't be a small project but there are other ways to accomplish it (we can discuss), which can take only a few seconds. There are some drawbacks though. For example, we could run:

`redocly bundle && redocly split` (I think).

jvc9109:
@adamaltman sure I will be happy to talk with you about the options for topic.

adamaltman (approved):

PR #956, opened 2022-08-08
Add extension for date formats

jvc9109:
add extension for properly format dob and simple dates in php-sdk

### DIFF
--- a/openapi/components/schemas/AML.yaml
+++ b/openapi/components/schemas/AML.yaml
@@ -119,6 +119,7 @@ properties:
       description: Date of birth.
       format: date
       example: 1706-01-17
+      x-simple-date-format: true
   aliases:
     type: array
     readOnly: true
--- a/openapi/components/schemas/Contact/ContactObject.yaml
+++ b/openapi/components/schemas/Contact/ContactObject.yaml
@@ -72,6 +72,7 @@ properties:
     format: date
     nullable: true
     example: 1980-04-01
+    x-simple-date-format: true
   jobTitle:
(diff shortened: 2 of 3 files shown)


justnero (approved):

openapi/components/schemas/AML.yaml
```diff
@@ -119,6 +119,7 @@ properties:
       description: Date of birth.
       format: date
       example: 1706-01-17
+      x-simple-date-format: true
```
ashkarpetin, line 122: What are the possible formats for `format: date`? I don't clearly understand the purpose of that extension.

jvc9109: There are two in the composition of the Models. A full datetime value like `2019-10-12T07:20:50.52Z` and a simple format like the Date of birth. This two formats can conflict in the json serialization of the php-sdk autogenerated.

ashkarpetin: Are not able to distinguish `format:date` and `format:date-time` during autogeneration?

justnero: Good point. @jvc9109 do we have a `date-time` set consistently in other places?

jvc9109: They both are mapped to `DatetimeImmutable` before template rendering or postprocessing. (Those steps are the ones we can have more control and customization)

jvc9109:
Closing PR to implement an alternative. Thanks @ashkarpetin  for the suggestions!

PR #980, opened 2022-08-16
refactor `ReadyToPayMethods` to discriminator

jvc9109:
refactor `ReadyToPayMethods` to use discriminator for sdk generation

### DIFF
--- a/openapi/components/schemas/ReadyToPay/ReadyToPayMethods.yaml
+++ b/openapi/components/schemas/ReadyToPay/ReadyToPayMethods.yaml
@@ -1,11 +1,15 @@
-type: array
-items:
-  anyOf:
+type: object
+discriminator: 
+  propertyName: method
+  x-enum-mapping:
     - $ref: ./Methods/ReadyToPayPaymentCardMethod.yaml
     - $ref: ./Methods/ReadyToPayAchMethod.yaml
     - $ref: ./Methods/ReadyToPayGenericMethod.yaml
     - $ref: ./Methods/ReadyToPayPayPalMethod.yaml
     - $ref: ./Methods/ReadyToPayKlarnaMethod.yaml
+properties: 
+  method:
+    type: string
 example:
   - method: payment-card
     feature:
--- a/openapi/components/schemas/ReadyToPay/ReadyToPayMethods.yaml
+++ b/openapi/components/schemas/ReadyToPay/ReadyToPayMethods.yaml
@@ -41,3 +41,4 @@ example:
     filters: []
   - method: Skrill
     filters: []
+x-no-abstract-methods: true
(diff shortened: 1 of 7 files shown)


openapi/components/schemas/ReadyToPay/ReadyToPayMethods.yaml
```diff
@@ -1,11 +1,15 @@
-type: array
-items:
-  anyOf:
+type: object
+discriminator: 
+  propertyName: method
+  x-enum-mapping:
```
ashkarpetin, line 4: Can we refactor our schema to not use that extension? I don't like it. It is not in compliance with OpenAPI, and hard to figure out when we should use it.

jvc9109: Since it includes a "Generic" option the result would look like:

```suggestion
    mapping:
       payment-card: ./Methods/ReadyToPayPaymentCardMethod.yaml
       ach: ./Methods/ReadyToPayAchMethod.yaml
      paypal: ./Methods/ReadyToPayPayPalMethod.yaml
      Klarna: ./Methods/ReadyToPayKlarnaMethod.yaml
      cash: ./Methods/ReadyToPayGenericMethod.yaml
      check: ./Methods/ReadyToPayGenericMethod.yaml
      AdvCash: ./Methods/ReadyToPayGenericMethod.yaml
      Alfa-click: ./Methods/ReadyToPayGenericMethod.yaml
      Alipay: ./Methods/ReadyToPayGenericMethod.yaml
      AstroPay Card: ./Methods/ReadyToPayGenericMethod.yaml
      AstroPay-GO: ./Methods/ReadyToPayGenericMethod.yaml
      BankReferenced: ./Methods/ReadyToPayGenericMethod.yaml
      bank-transfer: ./Methods/ReadyToPayGenericMethod.yaml
      bank-transfer-2: ./Methods/ReadyToPayGenericMethod.yaml
      ...
```
and so on for the 142 generic payment methods. It is a difficult compromise but we can explore an alternative for next cycle to change the way we handle this special cases

justnero: As an extension author, I should explain – it is designed to overcome duplicated enum declaration, like in this case with `ReadyToPayGenericMethod`, and should not be used for a 1-to-1 mapping.
Regarding OpenAPI compliance – it's actually pretty compliant as it is transformed into a plain `mapping` with all the lines in the [comment](https://github.com/Rebilly/api-definitions/pull/980#discussion_r946725998) above. The downside of doing it manually is at least the need to adjust several mappings each time an underlying enum gets changed.

justnero (approved):

justnero:
@Rebilly/php please review

justnero:
@Rebilly/js please review

ashkarpetin (approved):
Approved to unblock PR

PR #989, opened 2022-08-24
Add put and get notification settings

Boydbueno:


openapi/components/schemas/NotificationSettings.yaml
```diff
@@ -0,0 +1,19 @@
+type: object
+description: The settings for organization notifications.
+properties:
+  notifyOnUserAccessChanges:
+    description: Specifies if a notification should be sent when there are changes in user access.
+    type: boolean
+    default: false
+    example: false
+  notifyOnApiKeyAccessChanges:
+    description: Specifies if a notification should be sent when there are changes in API key access.
```
adamaltman, line 10: Does this include the creation of a new API key?

openapi/components/schemas/NotificationSettings.yaml
```diff
@@ -0,0 +1,19 @@
+type: object
+description: The settings for organization notifications.
+properties:
+  notifyOnUserAccessChanges:
+    description: Specifies if a notification should be sent when there are changes in user access.
```
adamaltman, line 5: Does this include the creation of a new user?

plugins/products-bundler/mapping/Core.yaml
```diff
@@ -39,6 +39,7 @@ x-tagGroups:
       - Custom fields
       - Files
       - Search
+      - Email notifications
```
adamaltman, line 42: This doesn't seem part of Core but part of Users IMO.

adamaltman (changes requested):

openapi/components/schemas/NotificationSettings.yaml
```diff
@@ -0,0 +1,19 @@
...
+    type: boolean
+    default: false
```
adamaltman, line 7: Why a default value? Also, shouldn't these be on by default with the default value in the email addresses list as the org creator (original owner)?

PR #1049, opened 2022-09-27
adjust server timestamp description

adamaltman:
## Summary

Any reference to "back-end" seems strange. What is "back-end" mean to external users?

Also, corrects the format description.

## Links

(Observed usage in recent PRs)

## Checklist

- [x] Writing style
- [x] API design standards

### DIFF
--- a/openapi/components/schemas/ServerTimestamp.yaml
+++ b/openapi/components/schemas/ServerTimestamp.yaml
@@ -1,5 +1,4 @@
 type: string
-description: Read-only timestamp. This value is automatically assigned
-  on back-end.
+description: Read-only timestamp in ISO 8601 date-time format. This value is automatically assigned.
 format: date-time
 readOnly: true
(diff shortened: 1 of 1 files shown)


nanobash (dismissed):

adamaltman:
I'm also not sure the "This value is automatically assigned" adds any value, as it is a read only timestamp and that implies that it is automatically assigned. So along these lines, there could be a lot of superfluous stuff added to this description -- because it is accurate doesn't mean it should be there.

openapi/components/schemas/ServerTimestamp.yaml
```diff
@@ -1,5 +1,4 @@
 type: string
-description: Read-only timestamp. This value is automatically assigned
-  on back-end.
+description: Read-only timestamp in ISO 8601 date-time format. This value is automatically assigned.
```
adamaltman, line 2:
```suggestion
description: Read-only timestamp in ISO 8601 date-time format.
```

n-fasano (approved):

nanobash (approved):

PR #1137, opened 2022-11-19
Timestamp when instead of at which

adamaltman:
## Summary

"Date and time when" seems more natural than "Date and time at which".

Thoughts?

## Checklist

- [ ] Writing style
- [ ] API design standards

### DIFF
--- a/WRITING-STYLE.md
+++ b/WRITING-STYLE.md
@@ -243,14 +243,14 @@ Coupon created time:
 
 ```yaml
   createdTime:
-    description: Date and time at which the coupon is created.
+    description: Date and time when the coupon is created.
 ```
 
 Coupon expired time:
 
 ```yaml
   expiredTime:
-    description: Date and time at which the coupon expires.
+    description: Date and time when the coupon expires.
 ```
 
 ### Boolean fields


XaosSintez (approved):
Agree

creocoder (approved):
LGTM, left alternate suggestions and some thoughts

WRITING-STYLE.md
```diff
@@ -243,14 +243,14 @@ Coupon created time:
 
 ```yaml
   createdTime:
-    description: Date and time at which the coupon is created.
+    description: Date and time when the coupon is created.
```
creocoder, line 246:
```suggestion
    description: Date and time the coupon was created.
```

WRITING-STYLE.md
```diff
@@ -243,14 +243,14 @@ Coupon created time:
...
 ```
 
 Coupon expired time:
 
 ```yaml
   expiredTime:
-    description: Date and time at which the coupon expires.
+    description: Date and time when the coupon expires.
```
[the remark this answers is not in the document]
creocoder: I also think original property `expiredTime` carries the wrong meaning. The name of the property implies the past tense. We should generally reconsider the naming of properties of this type. Possible good name: `expirationSheduledAt`.

WRITING-STYLE.md
creocoder: Reason: the shortest possible definition, without loss of meaning.

adamaltman: This suggestion violates the future-past rule.

ashkarpetin (approved):

WRITING-STYLE.md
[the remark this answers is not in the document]
adamaltman: Property names for date-time are an issue. But unless we version or deprecate it is going to be difficult to change property names.

WRITING-STYLE.md
creocoder: Got it, thank you

vykhrystiuk (approved):

noon-dawg (approved):

noon-dawg:
Good point. `at which` seems like a much older and more formal way of saying it 😄

PR #1143, opened 2022-11-22
Update contributing guide and corresponding rules

adamaltman:
## Summary

- Tune up that contributing guide.
- Enable more rules.
- Fix media type definitions for file downloads.

## Links

n/a

## Checklist

- [x] Writing style
- [x] API design standards

### DIFF
--- a/.redocly.lint-ignore.yaml
+++ b/.redocly.lint-ignore.yaml
@@ -1,0 +1,34 @@
+# This file instructs Redocly's linter to ignore the rules contained for specific parts of your API.
+# See https://redoc.ly/docs/cli/ for more information.
+openapi/paths/authentication-options.yaml:
+  no-http-verbs-in-paths:
+    - '#/'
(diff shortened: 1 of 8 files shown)


.redocly.lint-ignore.yaml
```diff
@@ -0,0 +1,34 @@
+# This file instructs Redocly's linter to ignore the rules contained for specific parts of your API.
+# See https://redoc.ly/docs/cli/ for more information.
+openapi/paths/authentication-options.yaml:
+  no-http-verbs-in-paths:
```
ashkarpetin, line 4: I am curios what is wrong with `/authentication-options` that no-http-verbs-in-paths is added?

adamaltman: "option" is an http verb (like post, put, patch, option is also an http verb) so we need to explicitly ignore it.

ashkarpetin (approved):

PR #1172, opened 2022-12-01
Rework payouts, plans, and previews

noon-dawg:
## Summary

This pull request reworks payouts, plans, and preview descriptions, and related files to conform to the [writing style](https://github.com/Rebilly/api-definitions/blob/main/WRITING-STYLE.md).

## Links

N/A

## Checklist

- [x] Writing style
- [x] API design standards


openapi/components/schemas/Common/CommonPlan.yaml
```diff
@@ -37,43 +42,46 @@ properties:
     allOf:
       - $ref: ../CurrencyCode.yaml
   currencySign:
-    description: Currency sign.
+    description: Currency symbol.
```
adamaltman, line 45: We decided to use the term sign over symbol though throughout the company. Why change that in the description?

openapi/components/schemas/Plans/Formulas/volume.yaml
```diff
@@ -6,35 +6,45 @@ allOf:
...
+          For example, based on the table below,
+          if a customer buys 6 units, they pay $3 per unit.
+          If a customer buys 1 unit, they pay $5.
 
-          Price per apple | Max quantity | Description
-          ----------------|--------------|------------
-          $5              | 1            | 1
-          $4              | 5            | 2 to 5
-          $3              | null         | 6 or more
+          Price per apple |Quantity range|
```
adamaltman, line 16: The old table has a column "Max quantity" which is indicating how the API expects data. There is a corresponding field `maxQuantity` in the API which expects that same value. Whereas, in this table there is a "Quantity range" but that isn't how the API works. Wouldn't it be important to explain these things?

noon-dawg: Thanks for pointing this out. The 'Max quantity' heading is important here, I'm not sure how I managed to overlook this

noon-dawg:
@Rebilly/php please review

openapi/components/schemas/Common/CommonPlan.yaml
```diff
@@ -45,36 +50,39 @@ properties:
...
+      Number of times the plan is modified.
+      Compare this value with materialized subscription item revision values.
   createdTime:
     $ref: ../CreatedTime.yaml
   updatedTime:
     $ref: ../UpdatedTime.yaml
   isTrialOnly:
     type: boolean
-    description: Whether a plan has a trial without recurring instructions.
+    description: Specifies whether a plan is a trial that does not have recurring instructions.
```
Boydbueno, line 81: I noticed that `whether` is used quite a lot instead of `if`. Isn't `whether` just a fancier word for the same thing?

openapi/components/schemas/Plans/Formulas/fixed-fee.yaml
```diff
@@ -6,9 +6,12 @@ allOf:
     properties:
       price:
         description: |-
-          For the very simple price when it's fixed and does not depend on the quantity chosen by customer.
+          Use this formula to charge for subscriptions that involve the same price,
```
Boydbueno, line 9: nitpick: It looks like you're trying to avoid using 'fixed price', but I think this is actually more clear:
```suggestion
          Use this formula to charge for subscriptions that have a fixed price,
```

openapi/components/schemas/Plans/Formulas/tiered.yaml
```diff
@@ -6,36 +6,49 @@ allOf:
     properties:
       brackets:
         description: |-
-          Price brackets, along with the price formula,
-          is used to calculate the amount to charge for the product on this plan on the invoice.
+          Use this formula to charge for units that are sold within defined quantity ranges, or tiers.
```
Boydbueno, line 9: It's unclear to me what the difference is between the Volume and Tiered formula. Is the only difference increasing or decreasing prices for the brackets, or can both formula's handle both?

openapi/paths/previews@orders.yaml
```diff
@@ -6,7 +6,9 @@ post:
   summary: Preview an order
   operationId: PostPreviewOrder
   x-sdk-operation-name: order
-  description: Use order preview before making an actual order.
+  description: |-
+    Previews an order.
+    Use this operation to preview a daft order before making an order.
```
Boydbueno, line 11:
```suggestion
    Use this operation to preview a draft order before making an order.
```

openapi/components/schemas/Common/CommonPlan.yaml
noon-dawg: Excellent point! `if` and `whether` are interchangeable. I alway felt that was for two possibilities, but it seems that `if` can also be used!

I'll make a follow up pull request to update this in the style guide and across the API docs. Thanks for spotting this

> 'Whether' is a conjunction that usually starts a subordinate clause that expresses an indirect question involving two stated or implied possibilities or alternatives. if can also be used in this sense.

openapi/components/schemas/Plans/Formulas/tiered.yaml
noon-dawg: Thanks for pointing this out. These descriptions were not clear and really needed work. I found them quite difficult

Reworked in [b149f21](https://github.com/Rebilly/api-definitions/pull/1172/commits/b149f21cc8a55d3dc4280718f28a0b2e49b319ba)

Boydbueno (dismissed):

ashkarpetin (dismissed):

ashkarpetin (approved):

swilczynski (approved):

PR #1268, opened 2023-02-01
Add missing Storefront operation to pause an order

ashkarpetin:
## Summary
Add missing Storefront operation to pause an order
Rename GetEventRuleHistoryVersion operation to GetEventRuleHistory for consistency with API

## Checklist

- [x] Writing style
- [x] API design standards

### DIFF
--- a/openapi/components/requestBodies/storefront/PostOrderPause.yaml
+++ b/openapi/components/requestBodies/storefront/PostOrderPause.yaml
@@ -1,0 +1,50 @@
+content:
+  application/json:
+    schema:
+      type: object
+      properties:
+        description:
+          description: Description of the pause reason in free form.
+          type: string
+          maxLength: 255
+          nullable: true
+        effectiveTime:
+          description: |-
+            Date when the service period pauses.
+
+            This time should be later than the current time.
+            If this time is earlier then the current time, the current time is used.
+
+            If this field is omitted, this value defaults to the current time.
+          type: string
+          format: date-time
+          nullable: true
+        endTime:
+          description: |-
+            Time when the pause ends and the subscription resumes billing.
+
+            To resume a subscription from this point in time,
+            use the current time or an earlier time.
+            If `endTime` is earlier then the current time, the current time is used.
+            If this field is empty, the subscription is indefinitely paused.
+          type: string
+          format: date-time
+          nullable: true
+        timeRemaining:
+          description: |-
+            Amount of time between the pause end time and the renewal time in ISO-8601 durations format.
+            By default, this value is computed as the time between the pause and the renewal.
+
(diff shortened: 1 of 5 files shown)


openapi/components/requestBodies/storefront/PostOrderPause.yaml
```diff
@@ -0,0 +1,34 @@
...
+        description:
+          description: Description of the pause reason in free form.
+          type: string
+          maxLength: 255
+          nullable: true
+        effectiveTime:
+          description: |-
+            Date when the service period pauses.
+
+            This time should be later than the current time.
```
lexerom, line 15:
```suggestion
```
Avoid using 'should'. Not sure if we need to delete it, but without it meaning stays the same.

openapi/components/requestBodies/storefront/PostOrderPause.yaml
```diff
@@ -0,0 +1,34 @@
...
+      type: object
+      properties:
+        description:
+          description: Description of the pause reason in free form.
+          type: string
+          maxLength: 255
+          nullable: true
+        effectiveTime:
+          description: |-
+            Date when the service period pauses.
```
lexerom, line 13:
```suggestion
            Date and time when the service period pauses.
```

ashkarpetin: I copied this from SubscriptionPause schema

https://github.com/Rebilly/api-definitions/blob/f5710175ff73cb281089abc8e2269f9773d68d09/openapi/components/schemas/Subscription/SubscriptionPause.yaml#L40

openapi/components/requestBodies/storefront/PostOrderPause.yaml
ashkarpetin: @noon-dawg  If there is mistake we need to fix in both places

colinbird: > This time must be later than the current time.

?

ashkarpetin: I made the changes.

colinbird (approved):

lexerom (approved):

openapi/components/requestBodies/storefront/PostOrderPause.yaml
noon-dawg: We should avoid using 'should' when something is required. 'Should' may seem optional and may also sound passive.

I must have missed that usage in `[api-definitions/openapi/components/schemas/Subscription/SubscriptionPause.yaml](https://github.com/Rebilly/api-definitions/blob/f5710175ff73cb281089abc8e2269f9773d68d09/openapi/components/schemas/Subscription/SubscriptionPause.yaml#L40). I'll update it now. Sorry about that

PR #1272, opened 2023-02-03
Move resources to ULID with prefix

nanobash:
## Summary

[Preview](https://preview.redoc.ly/all-rebilly/move-resources-to-ulid%2Fbatch-1/)

## Checklist

- [x] Writing style
- [x] API design standards


openapi/components/schemas/Journal/JournalRecord.yaml
```diff
@@ -9,7 +9,10 @@ required:
...
 properties:
   id:
-    $ref: ../ResourceId.yaml
+    type: string
+    description: Unique resource ID.
+    maxLength: 50
+    example: jrn_rec_0YVCXV2HE5DBT89QV7RXSJEZQ4
   journalEntryId:
     description: ID of the journal entry.
     readOnly: true
```
lexerom, line 18: Mb add maxlength and example here too?

lexerom: Actually, invoiceId and InvoiceItemId also without examples

nanobash: But it is exactly, what was before `$ref: ../ResourceId.yaml`. Even in some examples I am adding `example` if not specified, that might be topic of discussion. But here ResourceId.yaml content is moved and example value is updated, nothing else.

nanobash: I'd would not modify other items rather than what's the scope of the PR.

ashkarpetin (approved):

lexerom (approved):

PR #1286, opened 2023-02-13
Add GatewayAccount setting to allow zero amount authorization for setup transactions

colinbird:
## Summary

Some gateways allow $0 authorizations.  Right now we do $1 instead.

We can change this for setup transactions.

Alternatively, we could add a new setup instruction "verify" that does a $0 auth

## Checklist

- [x] Writing style
- [x] API design standards

### DIFF
--- a/openapi/components/schemas/GatewayAccount.yaml
+++ b/openapi/components/schemas/GatewayAccount.yaml
@@ -336,7 +336,7 @@ properties:
         This is the default behavior.
   allowZeroAuthorization:
     description: |-
-      Allow a zero dollar authorization. This is not supported by all gateways, the default value for an authroization is $1.00.
+      Allow a zero dollar authorization. This is not supported by all gateways, the default value for an authorization is $1.00.
     type: boolean
     default: false
   createdTime:
--- a/openapi/components/schemas/GatewayAccount.yaml
+++ b/openapi/components/schemas/GatewayAccount.yaml
@@ -334,6 +334,11 @@ properties:
       do-nothing: |-
         Does nothing except return an approved `setup` transaction.
         This is the default behavior.
+  allowZeroAuthorization:
+    description: |-
+      Allow a zero dollar authorization. This is not supported by all gateways, the default value for an authroization is $1.00.
+    type: boolean
+    default: false
   createdTime:
     $ref: CreatedTime.yaml
   updatedTime:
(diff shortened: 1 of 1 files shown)


openapi/components/schemas/GatewayAccount.yaml
```diff
@@ -334,6 +334,11 @@ properties:
       do-nothing: |-
         Does nothing except return an approved `setup` transaction.
         This is the default behavior.
+  allowZeroAuthorization:
+    description: |-
+      Allow a zero dollar authorization. This is not supported by all gateways, the default value for an authroization is $1.00.
```
colinbird, line 339:
```suggestion
      Allow a zero dollar authorization. This is not supported by all gateways, the default value for an authorization is $1.00.
```

openapi/components/schemas/GatewayAccount.yaml
```diff
@@ -334,6 +334,11 @@ properties:
       do-nothing: |-
         Does nothing except return an approved `setup` transaction.
         This is the default behavior.
+  allowZeroAuthorization:
+    description: |-
+      Allow a zero dollar authorization. This is not supported by all gateways, the default value for an authorization is $1.00.
```
colinbird, line 339:
```suggestion
      Allow a zero amount for authorization. This is not supported by all gateways, the default value for an authorization is 1.00 in the transaction's currency.
```

openapi/components/schemas/GatewayAccount.yaml
```diff
@@ -334,6 +334,11 @@ properties:
       do-nothing: |-
         Does nothing except return an approved `setup` transaction.
         This is the default behavior.
+  allowZeroAuthorizationSetup:
+    description: |-
+      Allow a zero amount for authorization as part of a setup transaction. This is not supported by all gateways, the default value for an authorization is 1.00 in the transaction currency.
```
vykhrystiuk, line 339:
```suggestion
      Allow a zero amount for authorization as part of a setup transaction.
      This is not supported by all gateways, the default value for an authorization is 1.00 in the transaction currency.
```

adamaltman (changes requested):
This seems wrong. It seems like they should send a setup transaction where the gateway is configured to do `authorize` for `$0 USD` for example. This should cause a zero dollar verification.

colinbird:
Can you explain a bit more what is wrong?  

The currency of the authorize comes from the (parent) setup transaction.

Do you mean that this gateway configuration should not be only for setup transactions but for all authorize transactions?

adamaltman:
There should not be any user configuration here.

PR #1343, opened 2023-04-04
Add AML settings API

mikerebilly:
## Summary
Add AML settings API. For now it is `confidence` settings, but there are upcoming additions as well.

Open to other ideas.

An alternative idea I had was adding arrays of object entries for the different permutations, and I suppose if they don't have entries for every permutation, then the missing ones would remain the defaults.

## Links
N/A

## Checklist

- [x] Writing style
- [x] API design standards


openapi/openapi.yaml
```diff
@@ -611,6 +611,8 @@ paths:
     $ref: './paths/aml-checks@{id}.yaml'
   '/aml-checks/{id}/review':
     $ref: './paths/aml-checks@{id}@review.yaml'
+  '/aml-settings':
+    $ref: './paths/aml-settings.yaml'
```
ashkarpetin, line 615: We already have `/kyc-settings` endpoint. I wonder if it might lead to the same case we had with credentials hashes endpoint when we created separate type for each credential type? I wonder if could we have a combined endpoint for global settings?

type: kyc, aml

```
GET /settings/{type}
PUT /settings/{type}
```

mikerebilly: Maybe. I think it's ok as is, but it is a viable alternative. It would of course add scope to this and extra time. This is currently an API only ITS would be aware of (for now). Just something to consider.

ashkarpetin: It is only used by our apps. Just an idea.

mikerebilly: I think you're right. I'm wondering if I can introduce this API for ITS, then refactor it (along w/ KYC endpoint path) later. I would think it's only an issue or two, along with one BE PR and one or two FE PR's (SDK, Recomm and/or Workbench etc.)

ashkarpetin (approved):

noon-dawg (approved):

PR #1350, opened 2023-04-10
Add Directa24 setting to use API v3

dara123:
## Checklist

- [x] Writing style
- [x] API design standards

### DIFF
--- a/openapi/components/schemas/GatewayAccountConfig/Directa24.yaml
+++ b/openapi/components/schemas/GatewayAccountConfig/Directa24.yaml
@@ -56,3 +56,6 @@ allOf:
           skipStep:
             type: boolean
             description: Specifies whether to skip the enter user personal information step.
+          useV3Api:
+            type: boolean
+            description: Specifies whether to use Directa24 API version 3.


openapi/components/schemas/GatewayAccountConfig/Directa24.yaml
```diff
@@ -56,3 +56,6 @@ allOf:
           skipStep:
             type: boolean
             description: Specifies whether to skip the enter user personal information step.
+          useV3Api:
```
ashkarpetin, line 59: I wonder if we should use v3 API by default and add  `useLegacyApi` setting instead?

ashkarpetin (approved):

openapi/components/schemas/GatewayAccountConfig/Directa24.yaml
dara123: this will affect the current merchant's using Directa24. Maybe after we deploy this and test it on prod and we can inform our merchants to switch to v3 and we can make this setting true by default or we can even remove the current integration

justnero (approved):

noon-dawg (approved):

PR #1408, opened 2023-05-19
Fix nullable enum for sourceType, and nullable for taxCategoryId

swilczynski:
## Summary

> Note that null must be explicitly included in the list of enum values. Using nullable: true alone [is not enough](https://github.com/OAI/OpenAPI-Specification/blob/main/proposals/2019-10-31-Clarify-Nullable.md#if-a-schema-specifies-nullable-true-and-enum-1-2-3-does-that-schema-allow-null-values-see-1900) here.

## Checklist

- [x] Writing style
- [x] API design standards

### DIFF
--- a/openapi/components/schemas/Files/File.yaml
+++ b/openapi/components/schemas/Files/File.yaml
@@ -22,6 +22,7 @@ properties:
       - camera
       - organization-export
       - organization-closure-export
+      - null
     example: upload
   tags:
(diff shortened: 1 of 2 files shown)


swilczynski:
@Rebilly/php please review

ashkarpetin (approved):

Weetbix (approved):

PR #1413, opened 2023-05-24
Update schema for profile and membership

swilczynski:
## Checklist

- [x] Writing style
- [x] API design standards


openapi/components/schemas/Profile.yaml
```diff
@@ -29,11 +29,21 @@ properties:
     readOnly: true
     type: string
     nullable: true
+  permissions:
```
ashkarpetin, line 32: I think this should be readOnly

openapi/components/schemas/Profile.yaml
```diff
@@ -60,6 +76,14 @@ properties:
     type: string
     format: url
     deprecated: true
+  status:
+    description: Status of the user's account.
```
nanobash, line 80: I think here `'s` is not needed, or mistaken @noon-dawg?

```suggestion
    description: Status of the user account.
```

openapi/components/schemas/Profile.yaml
swilczynski: Thanks, I'll fix it

openapi/components/schemas/Membership.yaml
```diff
@@ -56,5 +61,14 @@ properties:
       For example, the user may be an organization admin.
     items:
       type: string
+  acl:
+    $ref: ./Acl.yaml
```
ashkarpetin, line 65:
```suggestion
    readOnly: true
    allOf:
      - $ref: ./Acl.yaml
```

ashkarpetin (approved):

noon-dawg (approved):

PR #1476, opened 2023-07-05
Reduce schema reuse and promote dependencies to top level directory

jvc9109:
## Summary
Cleaned subschemas (not referenced from the `paths` ) are moved to top level directory.

## Checklist

- [x] Writing style
- [x] API design standards


jvc9109:
Closing to refactor in a different way

PR #1485, opened 2023-07-07
Fix discriminator, nullable issues for gateway account resource

i-yasko:
## Summary
Fix discriminator, nullable issues.

The common fields were moved to separate file `BaseGatewayAccount.yaml` to avoid discriminator definition errors.

## Checklist

- [x] Writing style
- [x] API design standards


jvc9109 (changes requested):
Discriminator [was correct from OpenAPI perspective](https://redocly.com/docs/resources/discriminator/#allof-for-inheritance). Why is the change necessary, could you please expand in the description?. We are cleaning up the api-definitions to avoid the usage of `Common*` or `Base*` schemas. Those have no meaning as an API resource.

i-yasko:
The discriminator wasn't correct before these changes because of `oneOf` part was skipped. Please pay attention to changes this [file](openapi/components/schemas/GatewayAccount.yaml) and example of discriminator definition from [doc](https://redocly.com/docs/resources/discriminator/#when-to-use-the-swagger-discriminator).

The `BaseGatewayAccount.yaml` can be replaced by plain structure but it cause code duplication in a lot of places (all gateway account type files). I'm not sure, but maybe there is sense to use some base structure file for similar cases.

justnero:
@i-yasko `oneOf` is not a required part of discriminator. Please read the whole documentation page you are referring to. Section `allOf for inheritance` describes exactly what was implemented here. The change you are proposing is not more correct, not simpler to read.

jvc9109:
I agree with Arif. For the specific case of GatewayAccounts, the `allOf` ineheritance is the way to go for reusing properties that are common.

>it cause code duplication in a lot of places

This repo is about documentation, not coding. Duplication is expected, even encourage, if it helps to make each schema easy to read. We should not leak code architecture in the API docs design.

jvc9109 (approved):

ashkarpetin (approved):

justnero (approved):

PR #1503, opened 2023-07-18
Reduce schema reuse and promote dependencies to top level directory [Part 3]

jvc9109:
## Summary
The goal is to reduce excessive schema reuse in the scope of cleaning up the api-definitions files.
Cleaned subschemas (not referenced from the paths ) are moved to top level directory.

## Links
[Part 2](https://github.com/Rebilly/api-definitions/pull/1499)

## Checklist

- [x] Writing style
- [x] API design standards

PR #1508, opened 2023-07-19
Reduce schema reuse and promote dependencies to top level directory [Part 4]

jvc9109:
## Summary

## Links
[Part 3](https://github.com/Rebilly/api-definitions/pull/1503)

## Checklist

- [x] Writing style
- [x] API design standards


noon-dawg (approved):
Looks good on the docs side

2e3s (approved):

PR #1522, opened 2023-07-21
Reduce schema reuse and promote dependencies to top level directory [Part 6]

jvc9109:
## Summary
The goal is to reduce excessive schema reuse in the scope of cleaning up the api-definitions files.
Cleaned subschemas (not referenced from the paths ) are moved to top level directory.

## Links
[Part 5](https://github.com/Rebilly/api-definitions/pull/1511)

## Checklist

- [x] Writing style
- [x] API design standards

PR #1523, opened 2023-07-21
Reduce schema reuse and promote dependencies to top level directory [Part 7]

jvc9109:
## Summary

## Links
[Part 6](https://github.com/Rebilly/api-definitions/pull/1522)

## Checklist

- [x] Writing style
- [x] API design standards


ashkarpetin (approved):

noon-dawg (approved):
Looks good on the docs side

PR #1525, opened 2023-07-24
Reduce schema reuse and promote dependencies to top level directory [Part 8]

jvc9109:
## Summary

## Links
[Part 7](https://github.com/Rebilly/api-definitions/pull/1523)

## Checklist

- [x] Writing style
- [x] API design standards


openapi/components/requestBodies/PatchServiceCredential.yaml
```diff
@@ -2,8 +2,8 @@ content:
   application/json:
     schema:
       anyOf:
-        - $ref: ../schemas/Credentials/PatchCredential.yaml
-        - $ref: ../schemas/Credentials/PatchTaxJarCredential.yaml
-        - $ref: ../schemas/Credentials/PatchPlaidCredential.yaml
+        - $ref: ./PatchCredential.yaml
```
ashkarpetin, line 5:
```suggestion
        - $ref: ./PatchServiceCredential.yaml
```

I think we should rename it to avoid confusion which credential it is

ashkarpetin (dismissed):

openapi/components/requestBodies/PatchServiceCredential.yaml
jvc9109: I switch `PatchServiceCredential` to `PatchCredential` and viceversa. This way should be clear in UI which is being used

justnero (approved):

jvc9109:
@Rebilly/js  @Rebilly/docs  please review

lucasxciv (approved):

PR #1532, opened 2023-07-26
Simplify schemas directory tree [Part 1]

jvc9109:
## Summary
Last step in the cleaning up schemas task. The goal of this series of PRs is to move all schemas to the top level `openapi/schemas` directory, but `GatewayAccountConfig` will remain as the only subfolder.

## Checklist

- [ ] Writing style
- [ ] API design standards


adamaltman:
There are a lot of conflicts here. Is this still active?

PR #1534, opened 2023-07-26
Fix invoice spec

vykhrystiuk:
## Summary
Specify string length

## Checklist

- [x] Writing style
- [x] API design standards

### DIFF
--- a/openapi/components/schemas/Invoice.yaml
+++ b/openapi/components/schemas/Invoice.yaml
@@ -111,9 +111,13 @@ properties:
     nullable: true
     example: PO123456
     type: string
+    minimum: 0
+    maximum: 65535
   notes:
     description: Notes for the customer that are displayed on the invoice.
     type: string
+    minimum: 0
+    maximum: 50
   items:
     type: array
     description: Invoice items array.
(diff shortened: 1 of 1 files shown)


lexerom (dismissed):

openapi/components/schemas/Invoice.yaml
```diff
@@ -111,9 +111,13 @@ properties:
     nullable: true
     example: PO123456
     type: string
+    minimum: 0
+    maximum: 65535
```
nanobash, line 115: Why that specific length here and below?

vykhrystiuk: We have those restrictions, and they were not described in API docs.

nanobash (dismissed):

openapi/components/schemas/Invoice.yaml
vykhrystiuk: Actually it was incorrect, fixed

lexerom (approved):

nanobash (approved):

wmdmgv (approved):

PR #1562, opened 2023-08-23
Make more discriminators explicit

justnero:
## Summary
Most of these were intended as discriminators (single `enum` property to be distinguished by), but were not explicit about it.
[ReadyToPayKlarnaMethod](https://github.com/Rebilly/api-definitions/compare/explicit-discriminators?expand=1#diff-9907c040427975f67931a960a78d7cb2964f25ea71f488902ef46a0a62710c99) replaced with `oneOf` to be the same as PayPal feature.

## Checklist

- [x] Writing style
- [x] API design standards


noon-dawg (approved):
Looks good on the docs side

ashkarpetin (approved):

PR #1568, opened 2023-08-29
Add setting to Directa24

dara123:
## Checklist

- [x] Writing style
- [x] API design standards

### DIFF
--- a/openapi/components/schemas/GatewayAccountConfig/Directa24.yaml
+++ b/openapi/components/schemas/GatewayAccountConfig/Directa24.yaml
@@ -62,3 +62,6 @@ allOf:
           storeIdNumber:
             type: boolean
             description: Specifies whether to store the customer ID number for future use.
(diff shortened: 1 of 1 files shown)


openapi/components/schemas/GatewayAccountConfig/Directa24.yaml
```diff
@@ -62,3 +62,6 @@ allOf:
           storeIdNumber:
             type: boolean
             description: Specifies whether to store the customer ID number for future use.
+          roundDccAmount:
+            type: boolean
+            description: Specifies whether round the amount after DCC.
```
ashkarpetin, line 67: What is the round mode: half up or half down or round to even or round odd?

adamaltman: I agree, would be better to have a rounding mode for this in general as a general setting.

PR #1573, opened 2023-08-31
Fix tags definitions

swilczynski:
## Checklist

- [ ] Writing style
- [ ] API design standards


openapi/components/schemas/Tag.yaml
```diff
@@ -2,36 +2,20 @@ type: object
...
+      id:
+        description: ID of the tag.
+        allOf:
+          - $ref: ./ResourceId.yaml
+  - $ref: ./TagType.yaml
+  - $ref: ./TagName.yaml
+  - type: object
+    properties:
+      createdTime:
+        $ref: ./CreatedTime.yaml
```
adamaltman, line 17: This approach with CreatedTime is better than your approach of making TagType an object (it should be a string and used by $ref, not with allOf).

PR #1576, opened 2023-09-04
Add descriptions to Quote items

lexerom:
## Summary
Add `description` and `priceDescription` to quote items.

## Checklist

- [x] Writing style
- [x] API design standards


nanobash (dismissed):

tamasvigh (dismissed):

openapi/components/schemas/Quote.yaml
```diff
@@ -70,6 +70,20 @@ properties:
...
+          type: string
+          description: Description of the quote item.
+          example: Charge per approved transaction with DCC
+          default: ''
+          maxLength: 255
+        priceDescription:
+          type: string
+          description: |-
+            Price description of the quote item.
+            This value is used only with metered billing items.
```
noon-dawg, line 83: suggestion:: for readability
```suggestion
            This value is only used for metered billing items.
```

jvc9109 (dismissed):

noon-dawg (dismissed):

tamasvigh (dismissed):

mehicalen (dismissed):

giokaxo (dismissed):

openapi/components/schemas/Quote.yaml
```diff
@@ -140,10 +154,22 @@ properties:
...
             description:
               description: Description of the invoice item.
               type: string
-              maxLength: 1000
+              example: Charge per approved transaction with DCC
+              default: ''
+              maxLength: 255
+            priceDescription:
+              type: string
+              description: Price description of the invoice item.
```
adamaltman, line 169: Also true here?

>             This value is only used for metered billing items.

tamasvigh: Thanks! Updated in [6280d68](https://github.com/Rebilly/api-definitions/pull/1576/commits/6280d68145a1ff07ca559d3451965001b40fd6c0)

giokaxo (approved):

jvc9109 (approved):

noon-dawg (approved):

openapi/components/schemas/Quote.yaml
```diff
@@ -70,6 +70,20 @@ properties:
           oneOf:
             - $ref: ./Plans/OriginalPlan.yaml
             - $ref: ./Plans/FlexiblePlan.yaml
+        description:
+          type: string
+          description: Description of the quote item.
+          example: Charge per approved transaction with DCC
+          default: ''
```
ikram-patah, line 77: question:

For consistency, shouldn't this be `nullable` instead of with default empty string `''`? 
I don't see my properties using `default: ''`. Most of em have `nullable: true` instead.

openapi/components/schemas/Quote.yaml
```diff
@@ -140,10 +154,24 @@ properties:
...
+              example: Charge per approved transaction with DCC
+              default: ''
+              maxLength: 255
+            priceDescription:
+              type: string
+              description: |-
+                Price description of the invoice item.
+                This value is only used for metered billing items.
+              example: 50% of the markup for approved transactions
+              default: ''
```
ikram-patah, line 173: Same question as above regarding the nullable vs empty string.

openapi/components/schemas/Quote.yaml
lexerom: I see that as a theoretic question. I would use `nullable` for values where null is valid (never set, unknown value, etc) or no default value.

In our case we can use null thou, but it adds additional checks on the backend side.

PR #1593, opened 2023-09-13
Flatten KycRequestDocuments

justnero:
## Summary

This way, schema intuitively refers to a singular value object, instead of an array. This helps improve SDK generation.

## Checklist

- [x] Writing style
- [x] API design standards


openapi/components/schemas/KycRequest.yaml
```diff
@@ -13,7 +13,10 @@ properties:
   customerId:
     $ref: ./CustomerId.yaml
   documents:
-    $ref: ./KycRequestDocuments.yaml
+    type: array
```
ashkarpetin, line 16: Should we specifiy `minItems`?

justnero: Good point, added

ashkarpetin (approved):

lucasxciv (approved):

PR #1637, opened 2023-10-11
Add customerDocumentCustomField setting to dLocal

ashkarpetin:
## Checklist

- [x] Writing style
- [x] API design standards

### DIFF
--- a/openapi/components/schemas/GatewayAccountConfig/dLocal.yaml
+++ b/openapi/components/schemas/GatewayAccountConfig/dLocal.yaml
@@ -45,5 +45,5 @@ allOf:
           customerDocumentCustomField:
             type: string
             description: |-
-              Name of the transaction custom field to use for sending customer document.
+              Name of the custom field of transaction to use to send customer document.
               Use this setting to skip the intermediate step of collecting the customer document for payment card transactions.
--- a/openapi/components/schemas/GatewayAccountConfig/dLocal.yaml
+++ b/openapi/components/schemas/GatewayAccountConfig/dLocal.yaml
@@ -42,3 +42,8 @@ allOf:
             type: boolean
             default: false
             description: Specifies whether to create an installment plan and use it for payment.
+          customerDocumentCustomField:
+            type: string
+            description: |-
+              Name of the transaction custom field to use for sending customer document.
+              Use this setting to skip intermediate step of collecting customer document ID.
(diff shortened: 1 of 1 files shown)


dara123 (dismissed):

openapi/components/schemas/GatewayAccountConfig/dLocal.yaml
```diff
@@ -42,3 +42,8 @@ allOf:
             type: boolean
             default: false
             description: Specifies whether to create an installment plan and use it for payment.
+          customerDocumentCustomField:
+            type: string
+            description: |-
+              Name of the transaction custom field to use for sending customer document.
```
noon-dawg, line 48: suggestion: I think this reads better if we place the adjective before the object 
```suggestion
              Name of the custom transaction field to use to send the customer document.
```

openapi/components/schemas/GatewayAccountConfig/dLocal.yaml
```diff
@@ -42,3 +42,8 @@ allOf:
...
+              Use this setting to skip intermediate step of collecting customer document for payment card transactions.
```
noon-dawg, line 49: nitpick:
```suggestion
              Use this setting to skip the intermediate step of collecting customer the document for payment card transactions.
```

ashkarpetin: > of collecting customer the document 

That doesn't seem correct

openapi/components/schemas/GatewayAccountConfig/dLocal.yaml
ashkarpetin: That seems incorrect that's not a custom transaction field, that's a field which is taken from transaction.customFields

ashkarpetin: I think the semantics changes.

`Custom field of the transaction` (customFields is a property on transaction resource) vs `Custom transaction field` (which looks like any propery on transaction)

2e3s (approved):

colinbird (approved):

openapi/components/schemas/GatewayAccountConfig/dLocal.yaml
noon-dawg: I think your update on this description was perfect! Thanks for taking my input onboard 🙏

PR #1669, opened 2023-11-02
fix: remove extraneous API bundling because we have 'all'

adamaltman:
## Summary

Fix (indirectly) one remote content issue.

## Checklist

- [x] Writing style
- [x] API design standards


Weetbix (dismissed):

adamaltman:
pending readiness on the website side -- need to remove the old registry as "required" after that...

PR #1678, opened 2023-11-07
Add a concept of subscription stages

2e3s:
## Summary

Programming plans change, and somewhat replacing reactivation.

[Preview](https://preview.redoc.ly/all-rebilly/order-stages/tag/Orders/operation/PostSubscriptionStage)

## Checklist

- [ ] Writing style
- [ ] API design standards

### DIFF
--- a/openapi/components/schemas/SubscriptionStage.yaml
+++ b/openapi/components/schemas/SubscriptionStage.yaml
@@ -13,6 +13,24 @@ description: |-
   A new stage can be created for a canceled or a churned subscription.
   If the subscription is churned, it becomes "pending" until the start date as if it is recreated.
 properties:
+  subscriptionId:
+    description: ID of the subscription.
+    type: string
+    maxLength: 50
+    example: ord_01GYJPRKHBD6ZYHH897QCJMBS4
+  customerId:
+    readOnly: true
+    allOf:
+      - $ref: ./CustomerId.yaml
+  organizationId:
+    readOnly: true
+    allOf:
+      - $ref: ./OrganizationId.yaml
+  websiteId:
+    x-sortable: true
+    x-basic: true
+    allOf:
+      - $ref: ./WebsiteId.yaml
   number:
     description: Unique consecutive number of the stage.
     type: integer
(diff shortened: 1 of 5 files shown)


openapi/components/schemas/SubscriptionStage.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - "completed"
+      - "upcoming"
+    readOnly: true
+  fromServicePeriod:
+    description: |-
+      Number of service period when this subscription stage starts and the previous stage ends.
+      If this field is omitted, this value defaults to the next service period for the given subscription,
+      unless it conflicts with previous upcoming stages.
+    type: integer
+    minimum: 0
```
adamaltman, line 35: Is this better as an integer or a datetime?

openapi/paths/subscriptions@{id}@stages.yaml
```diff
@@ -0,0 +1,69 @@
+parameters:
+  - $ref: ../components/parameters/resourceId.yaml
+get:
+  x-products:
+    - Core
+  tags:
+    - Orders
+  summary: Retrieve the available order's stages
```
adamaltman, line 8:
```suggestion
  summary: Retrieve stages
```

openapi/components/schemas/SubscriptionStage.yaml
```diff
@@ -0,0 +1,44 @@
...
+  pendingActivation:
+    description: |-
+      Whether the stage is requires the initial invoice of the stage to be paid
```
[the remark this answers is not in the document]
adamaltman: What if someone wants to require all the previous invoices are paid to activate too?

openapi/components/schemas/SubscriptionStage.yaml
2e3s: It's much easier to be an integer for our implementation + client integrations, and it matches nearly all use cases for scheduling switches in the future.

openapi/components/schemas/SubscriptionStage.yaml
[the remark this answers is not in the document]
2e3s: This is "activation" concept developed onto a new level, so this field concerns only "initial" invoice requirement which is about the yet not existing invoice which is generated after this API call.
If this use case arises, we might have a field `pendingPreviousInvoicesPayment`, it would be a small task.

openapi/components/schemas/SubscriptionStage.yaml
2e3s: Subscription's `startTime` (which is more like "stage start time" in the greater picture) might be utilized to set the time when it starts from "pending" status, so it's not that we're losing any functionality here.

2e3s: I've removed or changed the fields which aren't relevant to the stage settings.

openapi/components/schemas/SubscriptionStage.yaml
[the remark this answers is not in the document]
adamaltman: It's a real current-day use case.

openapi/components/schemas/SubscriptionStage.yaml
```diff
@@ -0,0 +1,234 @@
...
+      - "upcoming"
+    readOnly: true
+  fromServicePeriod:
+    description: |-
+      Number of service period when this subscription stage starts and the previous stage ends.
+      If this field is omitted, this value defaults to the next service period for the given subscription,
+      unless it conflicts with previous upcoming stages.
+    type: integer
+    minimum: 0
+  pendingActivation:
```
adamaltman, line 56: stripe calls this payment_behavior, and it's not a boolean. Their approach is more future proofed. And it may address my other issue regarding payment previous invoices too.

noon-dawg (commented):
I added some minor text suggestions. Give me a shout if you have any questions!

openapi/components/schemas/SubscriptionStage.yaml
2e3s: Their subscription model looks different, but anyway, I've tried to shape it into `activationPolicy` with options "immediately, "after-initial-invoice-payment", "after-invoices-payment".

openapi/components/schemas/SubscriptionStage.yaml
```diff
@@ -0,0 +1,247 @@
+type: object
+description: |-
+  Subscription stage is a sequence of periods (up to infinite) of a subscription with the given settings
+  which can be followed by another stage with different subscription settings.
+  A stage change is similar to a new subscription creation, but with the same ID and following period number.
```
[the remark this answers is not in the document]
2e3s:
```suggestion
  A stage change is similar to the creation of a new subscription, but the stage uses the same subscription identifier along with the period number of the subsequent stage.
```

PR #1746, opened 2024-01-22
Add Invoice delinquencyTime

2e3s:
## Checklist

- [x] Writing style
- [x] API design standards

### DIFF
--- a/openapi/components/schemas/Invoice.yaml
+++ b/openapi/components/schemas/Invoice.yaml
@@ -335,12 +335,13 @@ properties:
       - $ref: ./OrganizationId.yaml
   delinquencyTime:
     description: |-
-      Date and time when the related order is considered delinquent and gets canceled,
-      which abandons this and other order's invoices with a status `unpaid` or `past-due`.
-      If it is `null`, this invoice won't make the related order delinquent.
-      If `dueTime` becomes higher than field's value, it becomes equal to `dueTime`.
+      Date and time when the related order is considered delinquent, and is canceled.
+      The current invoice is abandoned along with order invoices that have the `unpaid` or `past-due` status.
+      If this value is `null`, no delinquency time is configured.
+      If the `dueTime` of the order becomes greater than the `delinquencyTime` value, 
+      the `delinquencyTime` value becomes equal to `dueTime`.
 
-      If the invoice is not related to an order, this field can only be `null`.
+      If an invoice is not related to an order, this field can only be `null`.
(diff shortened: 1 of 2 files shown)


openapi/components/schemas/Invoice.yaml
```diff
@@ -333,6 +333,19 @@ properties:
...
     allOf:
       - $ref: ./OrganizationId.yaml
+  delinquencyTime:
+    description: |-
+      Date and time when the related order is considered delinquent and gets canceled,
+      which abandons this and other order's invoices with a status `unpaid` or `past-due`.
+      If it is `null`, this invoice won't make the related order delinquent.
+      If `dueTime` becomes higher than field's value, it becomes equal to `dueTime`.
+
+      If the invoice is not related to an order, this field can only be `null`.
```
noon-dawg, line 343: suggestion: for readability
```suggestion
      Date and time when the related order is considered delinquent, and is canceled.
      The current invoice is abandoned along with order invoices that have the `unpaid` or `past-due` status.
      If this value is `null`, no delinquency time is configured.
      If the `dueTime` of the order becomes greater than the `delinquencyTime` value, 
      the `delinquencyTime` value becomes equal to `dueTime`.

      If an invoice is not related to an order, this field can only be `null`.
```

openapi/components/schemas/Subscription.yaml
```diff
@@ -339,7 +339,10 @@ properties:
       - 'null'
     description: |-
       Length of time, in ISO-8601 durations format,
-      before which an order with an invoice in past the due state is automatically canceled.
+      which is added to the due time when setting the delinquency time of every order's invoice.
+      Reaching delinquency time of an invoice makes the order automatically canceled,
+      which abandons that and other order's invoices with a status `unpaid` or `past-due`.
```
noon-dawg, line 344: nitpick: [possessive apostrophes](https://github.com/Rebilly/api-definitions/blob/main/WRITING-STYLE.md#possessive-apostrophes)
```suggestion
      which is added to the due time of the order when setting the delinquency time for all related invoices.
      When the delinquency time of an invoice is reached, the order is automatically canceled. 
      The current invoice is abandoned along with order invoices that have the `unpaid` or `past-due` status.
```

noon-dawg (approved):
Looks good on the docs side

ashkarpetin (approved):

giokaxo (approved):

PR #1755, opened 2024-01-26
Fix auto-generation

dariashka:
## Checklist

- [ ] Writing style
- [ ] API design standards


openapi/paths/tags@{tag}@customers.yaml
```diff
@@ -69,7 +69,7 @@ delete:
                 type: string
   responses:
     '204':
-      description: Customers untagging process has been scheduled.
+      description: Customers untagging
```
noon-dawg, line 72: suggestion: for readability
```suggestion
      description: Untagging customers 
```

PR #1775, opened 2024-02-05
Add Storefront update usage limits

XaosSintez:
## Checklist

- [x] Writing style
- [x] API design standards


openapi/openapi.yaml
```diff
@@ -1359,6 +1359,8 @@ paths:
     $ref: paths/storefront/orders@{id}@cancellation.yaml
   '/storefront/orders/{id}/pause':
     $ref: paths/storefront/orders@{id}@pause.yaml
+  '/storefront/orders/{id}/usage-limits':
```
ashkarpetin, line 1362: I can be a part of StorefrontPatchOrder instead of separate operation. Am I missing something?

ashkarpetin: > And even if we will do so, we don't want to allow updating items using that endpoint - we need to use Quotes instead.
> `/storefront/orders/{id}/usage-limits` 

I don't see quote here

openapi/components/requestBodies/storefront/PostOrderUsageLimits.yaml
```diff
@@ -0,0 +1,78 @@
...
+                      quantity:
+                        type: integer
+                        description: Usage limit quantity.
+                        minimum: 1
+                      amount:
+                        type: number
+                        format: double
+                        description: Usage limit amount in the currency of the order.
+                        minimum: 0.01
+                  hardLimit:
```
ashkarpetin, line 51: User might be able to abuse that by changing limits to his own needs?

openapi/openapi.yaml
ashkarpetin: You can send `usageLimits` in PatchOrder payload. There is will be not difference, so it will depends if we need a separate permission for StorefrontPostUsageLimits. But I assume if StorefrontPatchOrder and StorefrontPostUsageLimits permission are both will available for billing portal JWT.

openapi/components/requestBodies/storefront/PostOrderUsageLimits.yaml
ashkarpetin: > Defines a limit where the customer can no longer use the service.
> Hard limits are used in addition to soft limits.
> When a soft limit is reached, a customer may receive a notification but the service can still be provided up to the hard limit value so that the customer can upgrade their plan.


If they set it too big, then they can use the service, and don't pay for it?

openapi/openapi.yaml
ashkarpetin: Same structure as in PostOrderUsageLimits, and to rename items to usageLimits.

openapi/components/requestBodies/storefront/PostOrderUsageLimits.yaml
lexerom: If hard limit reached - adding/editing usages by merchant will return 409 ("Unable to report usage once hard limit is reached."). 
So, merchant service must be provided only after reporting that usage (in case of success). Or check hard limit before providing it (if it takes some time - limits can be changed by customer and thus there can be issues with it).

ashkarpetin: > If hard limit reached - adding/editing usages by merchant will return 409 ("Unable to report usage once hard limit is reached.").

We are taling about customer ability to change limits

openapi/openapi.yaml
ashkarpetin: For example , `usageLimits.0.limits` and `usageLimits.0.itemId`

ashkarpetin: We already a have a lot of API operaions, so should think twice before adding a new one. If you, @VladimirRebilly and @lexerom think that it is the good idea, so be it.

ashkarpetin: What would be the difference sending `usageLimits` to StorefrontPatchOrder vs `items` to `StorefrontPostUsageLimits`. They both will update order resource with order payload in response.

openapi/components/requestBodies/storefront/PostOrderUsageLimits.yaml
ashkarpetin: Got it

ashkarpetin: I was thinking of scenario when customer increases hard limits to use service, then abandons order, and doesn't pay for it. Then creates new order increases hard limits and uses service again, etc. I am not sure if this is realistic scenario.

openapi/openapi.yaml
lexerom: I am not sure if it is correct to add some new custom field to update existing resource in a different way than POST/PUT (I don't have concerns, I just don't know). If you think it is correct, then go for it.

lexerom (approved):

ashkarpetin (approved):

noon-dawg (approved):

PR #1795, opened 2024-02-16
Add isTrialConverted field to subscription

nanobash:
## Summary

[Preview](https://rebilly-website--rem-git-rem-01hnb1ejh7zd3eqe72m-4264c0.preview.redocly.app/catalog/all/Orders/GetSubscriptionCollection/#!t=response&c=200&path=0/isTrialConverted)

## Checklist

- [x] Writing style
- [x] API design standards


openapi/components/schemas/Subscription.yaml
```diff
@@ -109,6 +109,14 @@ properties:
       If this value is `true`, recurring settings are ignored.
     type: boolean
     default: false
+  isTrialConverted:
+    description: |-
+      Specifies if an order had a trial conversion.
+      Order considered to has a trial conversion if at least one successful payment made on the order after a trial.
```
noon-dawg, line 115: suggestion: to help make more concise
```suggestion
      Specifies if an order had trial conversion.
      Trial conversion is when at least one successful payment is made on the order after a trial.
```

openapi/components/schemas/Subscription.yaml
```diff
@@ -109,6 +109,14 @@ properties:
       If this value is `true`, recurring settings are ignored.
     type: boolean
     default: false
+  isTrialConverted:
+    description: |-
+      Specifies if an order had trial conversion.
+      Trial conversion is when at least one successful payment is made on the order after a trial.
+    type: boolean
+    readOnly: true
+    default: false
```
creocoder, line 118:
```suggestion
```

openapi/components/schemas/Subscription.yaml
```diff
@@ -109,6 +109,14 @@ properties:
       If this value is `true`, recurring settings are ignored.
     type: boolean
     default: false
+  isTrialConverted:
+    description: |-
+      Specifies if an order had trial conversion.
+      Trial conversion is when at least one successful payment is made on the order after a trial.
```
creocoder, line 115: I think it better replace `order` term there with `subscription order` or `subscription`.

openapi/components/schemas/Subscription.yaml
creocoder: There are no any default value. It always known was trial converted or not.

openapi/components/schemas/Subscription.yaml
nanobash: I thought about it, however did not do because with current implementation in `Subscription.yaml` file we treat `subscription` as an `order.` On the other hand, I think we can go with `subscription order` keep balance with old and new terminology.

nanobash: I think still we need to change it to `Subscription` as it refers to it.

Also according to `update orders docs` we are doing so.

cc: @noon-dawg

noon-dawg: In my opinion, we should use `subscription` in this description. `order` doesn't add to the the meaning, and may possibly confuse the reader.

suggested text:
`
Specifies if a subscription had trial conversion.
Trial conversion is when at least one successful payment is made on the subscription after a trial.
`

noon-dawg (approved):
Looks good on the docs side

creocoder (approved):

vykhrystiuk (approved):

i-yasko (approved):

PR #1799, opened 2024-02-20
Add pending status to Dispute

mikerebilly:
## Summary
Add `pending` status as a Dispute status.

This accounts for Disputes (eg. Adyen disputes) where a dispute's first chargeback has been decided, but the window for a possible second chargeback is still open. When that window closes, another dispute update happens where the final outcome statusis decided (eg. won/lost).

## Links
N/A

## Checklist

- [x] Writing style
- [x] API design standards

### DIFF
--- a/openapi/components/requestBodies/Dispute.yaml
+++ b/openapi/components/requestBodies/Dispute.yaml
@@ -225,6 +225,7 @@ content:
             - won
             - lost
             - unknown
+            - pending
         postedTime:
           description: Date and time when the dispute is posted.
           type: string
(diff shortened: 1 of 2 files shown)


openapi/components/requestBodies/Dispute.yaml
```diff
@@ -225,6 +225,7 @@ content:
             - won
             - lost
             - unknown
+            - pending
```
ashkarpetin, line 228: Why we can't use `unknown` status?

mikerebilly:
We don't need this after all.

PR #1830, opened 2024-03-15
Add StorefrontPostSubscriptionReactivation

lucasxciv:
## Summary

[Preview](https://rebilly-website--rem-git-rem-01hnb1ejh7zd3eqe72m-173e4d.preview.redocly.app/catalog/all/Storefront-orders/StorefrontPostSubscriptionReactivation/)

Add `StorefrontPostSubscriptionReactivation`.

## Checklist

- [x] Writing style
- [x] API design standards


openapi/openapi.yaml
```diff
@@ -1361,6 +1361,8 @@ paths:
     $ref: paths/storefront/orders@{id}@pause.yaml
   '/storefront/subscriptions/{id}/change-items':
     $ref: paths/storefront/subscriptions@{id}@change-items.yaml
+  '/storefront/subscriptions/{id}/reactivation':
```
adamaltman, line 1364:
```suggestion
  '/storefront/subscriptions/{id}/reactivations':
```

openapi/openapi.yaml
```diff
@@ -1361,6 +1361,8 @@ paths:
...
+    $ref: paths/storefront/subscriptions@{id}@reactivation.yaml
```
adamaltman, line 1365:
```suggestion
    $ref: paths/storefront/subscriptions@{id}@reactivations.yaml
```

openapi/paths/storefront/subscriptions@{id}@reactivation.yaml
adamaltman, line 1: It must accept a request body -- what if someone wants to reactivate with a different plan or quantity?

adamaltman (changes requested):

openapi/paths/storefront/subscriptions@{id}@reactivation.yaml
```diff
@@ -0,0 +1,34 @@
...
+  responses:
+    '201':
+      description: Quote for requested reactivation is created and issued.
+      headers:
+        Location:
+          $ref: ../../components/headers/Location.yaml
+      content:
+        application/json:
+          schema:
+            $ref: ../../components/schemas/StorefrontQuoteReactivateOrder.yaml
```
adamaltman, line 26: I haven't checked what this schema is yet. To me, it implies that this is incorrectly structured paths -- reactivation shouldn't be a nested subresource.

https://redocly.com/blog/no-subresources/

openapi/openapi.yaml
```diff
@@ -1361,6 +1361,8 @@ paths:
     $ref: paths/storefront/orders@{id}@pause.yaml
   '/storefront/subscriptions/{id}/change-items':
     $ref: paths/storefront/subscriptions@{id}@change-items.yaml
+  '/storefront/subscriptions/{id}/reactivations':
```
adamaltman, line 1364: This is a subresource. Does the reactivation have any ID? Kind of -- it's a quote ID.
```suggestion
  '/storefront/quotes/{id}/':
```

Or `/storefront/reactivations/{id}/`

openapi/components/requestBodies/storefront/PostSubscriptionReactivation.yaml
```diff
@@ -0,0 +1,28 @@
+type: object
+description: Reactivate customer subscription.
+properties:
+  items:
+    type: array
```
adamaltman, line 5:
```suggestion
  items:
    orderId:
      type: string
      description: ID of the order associated with the reactivation quote.
    type: array
```

openapi/components/requestBodies/storefront/PostSubscriptionReactivation.yaml
```diff
@@ -0,0 +1,34 @@
+type: object
+description: Reactivate customer subscription.
+required:
+  - subscriptionId
+properties:
+  subscriptionId:
```
adamaltman, line 6: Are they reactivating a subscription or an order?

lucasxciv: It should be subscription (discussed internally).

adamaltman (approved):

noon-dawg (approved):

justnero (approved):

i-yasko (approved):

creocoder (changes requested):
It doesn't look correct that the subscription level endpoint requires `orderId`.

openapi/components/requestBodies/storefront/PostSubscriptionReactivation.yaml
```diff
@@ -0,0 +1,34 @@
+type: object
+description: Reactivate customer subscription.
+required:
+  - orderId
+properties:
+  orderId:
+    type: string
+    description: ID of the order associated with the reactivation quote.
+    example: ord_01HRF27SATGE4Z6PBJE6PD8328
```
creocoder, line 9:
```suggestion
  subscriptionId:
    type: string
    description: ID of the subscription associated with the reactivation quote.
    example: ord_01HRF27SATGE4Z6PBJE6PD8328
```

adamaltman: It's an order level endpoint. The point is to reactivate the order. The word order and subscription have been used interchangeably for a few years which causes a problem with understanding this. The ID expected here is actually an order ID. Unless I'm misunderstanding something. I think the confusion is that prior to your project, 1 subscription ~ 1 order. You couldn't have 1 order with >1 subscription (but you could have >1 item including >1 subscription items).

So this is a real question as to what is getting reactivated? If it is reactivating the subscription only -- then this is the wrong endpoint and functionality. It doesn't meet the business use case. The business use case is related to order reactivation.

creocoder: > The point is to reactivate the order
If we talk about new orders, this is not an entirely correct statement. An order may include multiple subscriptions. We may want to reactivate one of them, several or all of them.

> So this is a real question as to what is getting reactivated?
The action context for reactivations is subscriptions, not order.

> The business use case is related to order reactivation.
This can be achieved by reactivating all order subscriptions.

creocoder: Let me explain the difference between `orderId` and `subscriptionId` in that particular case. With `orderId` we we are losing the ability to partially reactivate subscriptions. For example, if an order contains three subscriptions, and we want to reactivate only one of them. By using `subscriptionId` (one value or array) we get the opportunity to reactivate one, several or all order subscriptions.

creocoder: In general, if within the framework of this resource it is advisable to always completely reactivate all order subscriptions, then I think everything is good with `orderId` in that case.

creocoder (approved):

PR #1833, opened 2024-03-19
Add method to payout request allocation

colinbird:
## Checklist

- [x] Writing style
- [x] API design standards

### DIFF
--- a/openapi/components/schemas/PayoutRequest.yaml
+++ b/openapi/components/schemas/PayoutRequest.yaml
@@ -84,6 +84,8 @@ properties:
             - canceled
             - declined
             - unknown
+        method:
+          $ref: ./PaymentMethod.yaml
         gatewayName:
           description: Name of the payment gateway that processed, or is selected to process, the transaction.
           x-label: Gateway account


dara123 (approved):

i-yasko (approved):

openapi/components/schemas/PayoutRequest.yaml
```diff
@@ -84,6 +84,8 @@ properties:
             - canceled
             - declined
             - unknown
+        method:
```
ashkarpetin, line 87: Should it be called paymentMethod?

colinbird: I see we use both, but we use `method` more often, including in paymentinstrument

lexerom (approved):

openapi/components/schemas/PayoutRequest.yaml
ashkarpetin: We use it in context of paymentInstrument: paymentInstrument.method.

colinbird: Seems like a wider question, we also use method inside token.  I am not sure

ashkarpetin (approved):

nanobash (approved):

PR #1839, opened 2024-03-21
Add churnTimePolicy to subscription cancellation

i-yasko:
## Summary

The `churnTimePolicy` field adds simplicity to specify moment when subscription should be cancelled. 
If merchant wants to use specific date-time for subscription cancellation, `churnTime` allows to do that

## Checklist

- [x] Writing style
- [x] API design standards


openapi/components/schemas/SubscriptionCancellation.yaml
```diff
@@ -94,9 +94,11 @@ properties:
   updatedTime:
     $ref: ./UpdatedTime.yaml
   churnTime:
-    description: Date and time when the subscription is deactivated.
+    description: Specifies when the subscription is to be canceled.
```
2e3s, line 97: `Time` suffix has nearly universally time format in this API, it would better be `churnTimePolicy` or something with a different naming.
Also, it may be good to keep `churnTime` as a time anyway to be able to filter out and see subscriptions which are about to be churned.

i-yasko: I agree, it would be better to change name but in this case clients have to be change their integration, in order to handle new field

i-yasko: I think, we can add `churnTimePolicy` field to response in order to indicate which policy is used for cancellation. It also can be added to request as well, `churnTime` will be marked as deprecated in this case. What do you think?

2e3s: I think if make the change, it has to be in the request most importantly. And then it may be in the response.

openapi/components/schemas/SubscriptionCancellation.yaml
```diff
@@ -2,6 +2,7 @@ type: object
 required:
   - subscriptionId
   - churnTime
```
ikram-patah, line 4: Is `churnTime` still required if `churnTimePolicy` is provided?

i-yasko: No, one of these fields should present in the request. If both are presented, `churnTimePolicy ` takes precedence

openapi/components/schemas/SubscriptionCancellation.yaml
```diff
@@ -97,6 +97,17 @@ properties:
     description: Date and time when the subscription is deactivated.
     type: string
     format: date-time
+  churnTimePolicy:
+    description: |-
+      Specifies when the subscription is to be canceled.
+      The `churnTimePolicy` is takes precedence over the `churnTime` in request.
```
noon-dawg, line 103: nitpick: typo
```suggestion
      The `churnTimePolicy` takes precedence over the `churnTime` in request.
```

openapi/components/schemas/SubscriptionCancellation.yaml
```diff
@@ -97,6 +97,17 @@ properties:
     description: Date and time when the subscription is deactivated.
     type: string
     format: date-time
+  churnTimePolicy:
+    description: |-
+      Specifies when the subscription is to be canceled.
```
VladimirRebilly, line 102:
```suggestion
      Specifies when the subscription is to be churned.
```
or
```suggestion
      Specifies when the subscription is to be deactivated.
```
based on the description of `churnTime` field

VladimirRebilly (approved):

noon-dawg (approved):

ikram-patah (approved):

PR #1845, opened 2024-03-26
Fix GatewayAccount PUT SDK operation name

justnero:
## Summary

This was a mistake, as `update` is already used for `PATCH`.

## Checklist

- [x] Writing style
- [x] API design standards

### DIFF
--- a/openapi/paths/gateway-accounts@{id}.yaml
+++ b/openapi/paths/gateway-accounts@{id}.yaml
@@ -40,7 +40,7 @@ put:
     - Gateway accounts
   summary: Upsert a gateway account
   operationId: PutGatewayAccount
-  x-sdk-operation-name: update
+  x-sdk-operation-name: upsert
   description: Creates or updates (upserts) a gateway account with a specified ID.
   requestBody:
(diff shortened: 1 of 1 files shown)


nanobash (approved):

lexerom (approved):

openapi/paths/gateway-accounts@{id}.yaml
```diff
@@ -40,7 +40,7 @@ put:
     - Gateway accounts
   summary: Upsert a gateway account
   operationId: PutGatewayAccount
-  x-sdk-operation-name: update
+  x-sdk-operation-name: upsert
```
ashkarpetin, line 43: I see that we have 6 cases of `patch` x-sdk-operation-name for patch operations and 2 cases of upsert for put operations. I wonder if better to rename `x-sdk-operation-name: patch` for patch method?

justnero: Good catch. I'll investigate in it to find the least harmful and the best overall solution to this uniformity issue.

PR #1868, opened 2024-04-11
Fix filter and collection docs links

mikerebilly:
## Summary
Remove old documentation about filters and expand, and link to new docs. Also fixed `sort` definition and description.

## Checklist

- [x] Writing style
- [x] API design standards

PR #1870, opened 2024-04-11
Revert sort type change

mikerebilly:
## Summary
#1868 turned sort into `type: string` with `style: form`, the original definition was correct. I kept the description update though.

## Links
https://swagger.io/specification/#style-values:~:text=200.B%3D150-,form,false,-color%3D

## Checklist

- [x] Writing style
- [x] API design standards

### DIFF
--- a/openapi/components/parameters/collectionSort.yaml
+++ b/openapi/components/parameters/collectionSort.yaml
@@ -3,7 +3,6 @@ in: query
 description: |-
   Sorts and orders the collection of items. To sort in descending
   order, prefix with `-`. Multiple fields can be sorted by separating each with `,`.
-example: -createdTime,firstName
 style: form
 explode: false
 schema:
--- a/openapi/components/parameters/collectionSort.yaml
+++ b/openapi/components/parameters/collectionSort.yaml
@@ -7,4 +7,6 @@ example: -createdTime,firstName
 style: form
 explode: false
 schema:
-  type: string
+  type: array
+  items:
+    type: string


noon-dawg (approved):

mehicalen (approved):

openapi/components/parameters/collectionSort.yaml
```diff
@@ -3,8 +3,9 @@ in: query
 description: |-
   Sorts and orders the collection of items. To sort in descending
   order, prefix with `-`. Multiple fields can be sorted by separating each with `,`.
-example: -createdTime,firstName
 style: form
 explode: false
 schema:
-  type: string
+  type: array
```
ashkarpetin, line 9: So with array of string the query should look like `?sort[]=-createdTime&sort[]=firstName`? It doesn't seem corresponds to description.

ashkarpetin (approved):

openapi/components/parameters/collectionSort.yaml
ashkarpetin, line 9: I wonder why `collectionExpand` described different way 

https://github.com/Rebilly/api-definitions/blob/5fbaee835cd2faccd4c148c24a9567b98cc8357e/openapi/components/parameters/collectionExpand.yaml#L9-L10

justnero: That's a fair point. It's actually the same thing and should be defined the same way.

justnero (approved):

PR #1874, opened 2024-04-17
Refactor journal accounts proposal

jvc9109:
## Checklist

- [ ] Writing style
- [ ] API design standards


openapi/components/schemas/JournalAccountBalance.yaml
```diff
@@ -0,0 +1,32 @@
+type: object
+required:
+  - name
```
adamaltman, line 3: Name is not a property below.

PR #1875, opened 2024-04-19
Add abandon-scheduled-payments rule action

i-yasko:
## Summary

Added `abandon-scheduled-payments` rule action which allows to change default behaviour of invoice in case of reaching delinquency. 

If rule is configured, invoice change state from `past-due` to `abandoned`.

## Checklist

- [x] Writing style
- [x] API design standards


ashkarpetin (approved):

noon-dawg (approved):
Looks good on the docs side

giokaxo (approved):

PR #1878, opened 2024-04-24
docs: rate limiting

danio1024:
## Summary
We are adding rate limits on sandbox environment. This PR documents that mechanism. 
Since I've had to iterate over all endpoints I've also quoted all response codes so it is consistent among our definitions.
## Checklist

- [x] Writing style
- [x] API design standards

### DIFF
--- a/openapi/components/headers/X-RateLimit-Limit.yaml
+++ b/openapi/components/headers/X-RateLimit-Limit.yaml
@@ -1,4 +1,4 @@
-description: Overall rate limit tokens for this request.
+description: Total number of rate limit tokens for this request.
 schema:
   type: integer
 example: 3600
--- a/openapi/components/headers/X-RateLimit-Limit.yaml
+++ b/openapi/components/headers/X-RateLimit-Limit.yaml
@@ -1,0 +1,4 @@
+description: Overall rate limit tokens for this request.
+schema:
+  type: integer
+example: 3600
--- a/openapi/components/headers/X-RateLimit-Limit.yaml
+++ b/openapi/components/headers/X-RateLimit-Limit.yaml
@@ -1,4 +1,6 @@
-description: Total number of rate limit tokens for this request.
+description: |-
+  Total number of rate limit tokens for this request within a rate limit period.
+  For more information, see [Rate limits](#section/Rate-limits).
 schema:
   type: integer
 example: 3600
--- a/openapi/components/headers/X-RateLimit-Remaining.yaml
+++ b/openapi/components/headers/X-RateLimit-Remaining.yaml
@@ -1,4 +1,6 @@
-description: Remaining rate limit tokens for this request.
+description: |-
+  Remaining number of rate limit tokens for this request within the rate limit period. 
+  For example, in the sandbox environment, rate limits for non-GET endpoints are set at 3000 requests per 10 minutes.
 schema:
   type: integer
 example: 3600
--- a/openapi/components/headers/X-RateLimit-Remaining.yaml
+++ b/openapi/components/headers/X-RateLimit-Remaining.yaml
@@ -1,0 +1,4 @@
+description: Remaining rate limit tokens for this request.
+schema:
+  type: integer
+example: 3600
--- a/openapi/components/headers/X-RateLimit-Retry-After.yaml
+++ b/openapi/components/headers/X-RateLimit-Retry-After.yaml
@@ -1,4 +1,4 @@
-description: UTC timestamp after which the rate limit will reset and the request can be retried.
+description: UTC timestamp after which the rate limit resets and the request can be retried.
 schema:
   type: integer
 example: 1713187500
--- a/openapi/components/headers/X-RateLimit-Retry-After.yaml
+++ b/openapi/components/headers/X-RateLimit-Retry-After.yaml
@@ -1,4 +1,4 @@
-description: UTC timestamp after which the rate limit resets and the request can be retried.
+description: UTC timestamp after which rate limit resets and the request can be retried.
 schema:
   type: integer
 example: 1713187500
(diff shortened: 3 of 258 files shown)


openapi/components/headers/X-RateLimit-Limit.yaml
```diff
@@ -0,0 +1,4 @@
+description: Overall rate limit tokens for this request.
```
noon-dawg, line 1: suggestion: I'm not clear on what 'overall' means in this description

I think it's the total number?
```suggestion
description: Total number of rate limit tokens for this request.
```

openapi/components/headers/X-RateLimit-Retry-After.yaml
```diff
@@ -0,0 +1,4 @@
+description: UTC timestamp after which rate limit will reset and the request can be retried.
```
noon-dawg, line 1: nitpick: typo
```suggestion
description: UTC timestamp after which the rate limit will reset and the request can be retried.
```

openapi/description.md
```diff
@@ -71,3 +71,12 @@ If you would like to implement a particular use case,
...
 see [Get started](https://www.rebilly.com/docs/dev-docs/get-started/).
+
+# Rate limits
+
+Rebilly enforces rate limits on the API to ensure that no single organization can consume too many resources.
+Rate limits are applied to the organization, and not to the API key.
+Currently, rate limits are enforced for non-GET endpoints in sandbox environment and they are set to 3000 requests per 10 minutes.
+Exact numbers of consumed requests can be found in the `X-RateLimit-Limit` and `X-RateLimit-Remaining` headers in the response.
+In the event that the rate limit is exceeded, the API returns a `429 Too Many Requests` response
+and a `X-RateLimit-Retry-After` header with the UTC timestamp when the rate limit resets/replenishes.
```
noon-dawg, line 82: suggestion: for readability
```suggestion
Rebilly enforces rate limits on the API to ensure that no single organization consumes too many resources.
Rate limits are applied to the organization, and not to the API key.
In sandbox environment, rate limits are enforced for non-GET endpoints and are set at 3000 requests per 10 minutes.
You can find the exact number of consumed requests in the `X-RateLimit-Limit` and `X-RateLimit-Remaining` headers in the response.
If the rate limit is exceeded, the API returns a `429 Too Many Requests` response
and a `X-RateLimit-Retry-After` header that includes a UTC timestamp of when the rate limit resets.
```

openapi/components/headers/X-RateLimit-Limit.yaml
```diff
@@ -0,0 +1,4 @@
+description: Total number of rate limit tokens for this request.
```
noon-dawg, line 1: suggestion: to add more detail

I think it would be good to point to the Rate limit section to provide more information 

```suggestion
description: |-
  Total number of rate limit tokens for this request.
  For more information, see [Rate limits](#section/Rate-limits).
```
It may also be good to add `rate limit period` to the description 
```suggestion
description: |-
  Total number of rate limit tokens for this request within a rate limit period.
  For more information, see [Rate limits](#section/Rate-limits).
```

openapi/components/headers/X-RateLimit-Remaining.yaml
```diff
@@ -0,0 +1,4 @@
+description: Remaining rate limit tokens for this request.
```
noon-dawg, line 1: suggestion: to add more detail

I think it would be good to add `within the rate limit period`

```suggestion
description: Remaining number of rate limit tokens for this request within the rate limit period. 
```
and possibly an example 
```suggestion
description: |-
  Remaining number of rate limit tokens for this request within the rate limit period. 
  For example, in the sandbox environment, rate limits for non-GET endpoints are set at 3000 requests per 10 minutes.
```

openapi/components/headers/X-RateLimit-Retry-After.yaml
```diff
@@ -0,0 +1,4 @@
+description: UTC timestamp after which the rate limit will reset and the request can be retried.
```
noon-dawg, line 1: suggestion: to avoid future tense
```suggestion
description: UTC timestamp after which the rate limit resets and the request can be retried.
```

openapi/components/schemas/ProblemTypes/TooManyRequests.yaml
```diff
@@ -0,0 +1,9 @@
+allOf:
+  - type: object
+    properties:
+      status:
+        type: integer
+        description: HTTP status code.
+        minimum: 429
+        maximum: 429
+  - $ref: ./BaseProblem.yaml
```
adamaltman, line 9: Better to consolidate this into a single schema that doesn't use `allOf`.

Why?

We can provide better examples for each of the properties (and the reason we do this separate definition for only the status property is to provide a further constraint on the value of the status -- the same could probably be done for the `type` and the `title` for the **Too many requests**).

openapi/components/headers/X-RateLimit-Retry-After.yaml
adamaltman, line 1: Is this field referenced anywhere?

danio1024: It was supposed to be added to the 429 error response as you've mentioned in other comment, it now should be referenced.

openapi/components/schemas/ProblemTypes/TooManyRequests.yaml
```diff
@@ -0,0 +1,34 @@
+type: object
+additionalProperties: true
+minProperties: 1
+description: RFC-7807 [problem details](https://tools.ietf.org/html/rfc7807) JSON object.
```
adamaltman, line 4:
```suggestion
description: RFC-9457 [problem details](https://datatracker.ietf.org/doc/html/rfc9457) JSON object.
```
The RFC was superseded by 9457.

adamaltman (approved):
Nice. 🎉 

I have a minor change request (we need to change the other problems to RFC 9457 too though.

noon-dawg (approved):
Looks good on the docs side

vykhrystiuk (approved):

openapi/components/schemas/ProblemTypes/TooManyRequests.yaml
danio1024: @adamaltman created an issue to fix that in all problems.

PR #1885, opened 2024-05-03
Implement subscription.trialPeriod field

nanobash:
## Summary

[Preview](https://rebilly-website--rem-git-rem-01hnb1ejh7zd3eqe72m-251e52.preview.redocly.app/catalog/all/orders/getsubscriptioncollection#orders/getsubscriptioncollection/t=response&c=200&path=&d=0/trial/period)

- [x] Implement `Subscription.trialPeriod` field

## Checklist

- [x] Writing style
- [x] API design standards

### DIFF
--- a/openapi/components/schemas/Subscription.yaml
+++ b/openapi/components/schemas/Subscription.yaml
@@ -102,7 +102,6 @@ properties:
           Plans without trial prices are free trials.
         type: boolean
       endTime:
-        deprecated: true
         description: |-
           Time and date when the trial ends.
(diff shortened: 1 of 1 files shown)


openapi/components/schemas/Subscription.yaml
```diff
@@ -102,13 +102,25 @@ properties:
           Plans without trial prices are free trials.
         type: boolean
       endTime:
+        deprecated: true
```
nanobash, line 105: I think we can deprecate it later with separate PR.

